PR #3, opened 2015-11-18
VALS project - Audaspace 1.2

DethonUSAL:
Functionalities added: Random sounds, dynamic music, playback manager, convolution, and HRTFs.


.gitignore
```diff
@@ -0,0 +1,189 @@
+## Ignore Visual Studio temporary files, build results, and
```
neXyon: There shouldn't be a .gitignore file in the repository, especially not for build files. I recommend you do out-of-source builds! If you need/want a .gitignore file, add one to the clone directory and add .gitignore in the file, so that it doesn't get committed.

AUTHORS
```diff
@@ -1,4 +1,4 @@
-Main author: Jörg Müller <nexyon@gmail.com>
```
neXyon: I cannot see the difference between the two fields here, maybe it's because there is a BOM added or removed? Please make sure it's consistent throughout the repository, I also saw files below having the issue.

CMakeLists.txt
```diff
@@ -560,9 +605,9 @@ endif()
 list(APPEND SRC ${CMAKE_CURRENT_BINARY_DIR}/PluginManager.cpp)
 
 # directories
-
-include_directories(${INCLUDE})
-link_directories()
```
neXyon: Why did you delete these two lines? Are they not necessary?

CMakeLists.txt
```diff
@@ -576,9 +621,9 @@ set(LIB_DESTINATION "lib${LIB_SUFFIX}")
 
 # library
 
+
 add_library(audaspace ${LIBRARY_TYPE} ${SRC} ${HDR})
-target_link_libraries(audaspace ${LIBRARIES})
-set_target_properties(audaspace PROPERTIES SOVERSION ${AUDASPACE_VERSION})
```
neXyon: You add FFTW_LIBRARY to LIBRARIES, this should be enough for ${LIBRARIES} to include the fftw library, no? Also please keep the SOVERSION property.

CMakeLists.txt
```diff
@@ -654,7 +699,7 @@ endif()
 if(BUILD_DEMOS)
 	include_directories(${INCLUDE})
 
-	set(DEMOS audaplay audaconvert audaremap signalgen)
+	set(DEMOS audaplay audaconvert audaremap signalgen randsounds)
```
neXyon: You forgot to add your other demos here.

CMakeLists.txt
```diff
@@ -690,14 +750,14 @@ endif()
...
-			ARCHIVE DESTINATION ${LIB_DESTINATION}
-		)
+	add_library(audaspace-c ${LIBRARY_TYPE} ${C_SRC} ${C_HDR})
+	target_link_libraries(audaspace-c audaspace)
+	set_target_properties(audaspace-c PROPERTIES SOVERSION ${AUDASPACE_VERSION})
+	install(TARGETS audaspace-c
+		RUNTIME DESTINATION ${BIN_DESTINATION}
+		LIBRARY DESTINATION ${LIB_DESTINATION}
+		ARCHIVE DESTINATION ${LIB_DESTINATION}
+	)
```
neXyon: Indentation change?

bindings/C/AUD_Sound.cpp
```diff
@@ -621,3 +623,45 @@ AUD_API AUD_Sound* AUD_Sound_copy(AUD_Sound* sound)
...
+	{
+		return new AUD_Sound(new SoundList(random));
+	}
+	catch (Exception&)
+	{
+		return nullptr;
+	}
+}
+
+AUD_API int AUD_Sound_list_addSound(AUD_Sound* list, AUD_Sound* sound)
```
neXyon: I think AUD_SoundList_addSound would be a better name.

bindings/python/PyAPI.cpp
```diff
@@ -77,7 +79,13 @@ PyInit_aud()
 	if(!initializeSequenceEntry())
 		return nullptr;
 
-	if(!initializeSequence())
+	if (!initializeSequence())
```
neXyon: Another consistency comment: let's keep the coding style in audaspace. Can you remove the space of all "if (", "for (", etc. => "if(", "for(" please!

bindings/python/PySound.cpp
```diff
@@ -1600,9 +1699,18 @@ static PyMethodDef Sound_methods[] = {
 	{"mix", (PyCFunction)Sound_mix, METH_O,
 	 M_aud_Sound_mix_doc
 	},
-	{"pingpong", (PyCFunction)Sound_pingpong, METH_NOARGS,
```
neXyon: The oddities you see in diffs. xD

demos/binaural.cpp
```diff
@@ -0,0 +1,208 @@
...
+		if (x >= 360)
+			x = 0;
+		source->setAzimuth(x);
+		//source->setElevation(y);
+	}
+
+	return 0;
+}
+
+void loadHRTFs(std::shared_ptr<HRTF> hrtfs)
```
neXyon: I'm sure you can implement this function a lot nicer! :-)

demos/convolution.cpp
```diff
@@ -0,0 +1,66 @@
...
+#include "devices/DeviceManager.h"
+#include "devices/IDevice.h"
+#include "devices/IDeviceFactory.h"
+#include "devices/IHandle.h"
+#include "plugin/PluginManager.h"
+#include "file/File.h"
+#include "IReader.h"
+
+#include <iostream>
+#include <condition_variable>
```
neXyon: You don't need condition variables in this demo (and others). Please try to keep the number of includes as small as possible everywhere.

AUTHORS
DethonUSAL: Yeah, it's an added BOM. I'll delete it from all the files

CMakeLists.txt
DethonUSAL: I can't remember removing them, so it was most likely by mistake... I didn’t experience any problems though, probably because the include_directories() line is also called in the demos section.

demos/binaural.cpp
DethonUSAL: Yeah, that function is quite embarrassing. I wanted a "quick", platform independent way to load all the HRTF files. I'll make it so it loads all the files from one directory, though I won't be able to test the Unix version right away.

DethonUSAL:
I think this addressed all your comments, but I haven't been able to test the Unix versión of the binaural demo yet.

demos/binauralUnix.cpp
```diff
@@ -0,0 +1,126 @@
...
+	{
+		std::string filename = entry->d_name;
+		
+		if(filename.front() == 'R')
+		{
+			elev = std::stof(filename.substr(1, filename.find("e") - 1));
+			azim = std::stof(filename.substr(filename.find("e") + 1, filename.find("a") - filename.find("e") - 1));
+			hrtfs->addImpulseResponse(std::make_shared<StreamBuffer>(std::make_shared<File>(path + "/" + filename)), azim, elev);
+		}
+		found_file = FindNextFile(dir, &entry);
```
neXyon: this line needs to be removed

DethonUSAL: Oops, I forgot that one.

neXyon:
I just tried the binaural demo on linux, it works as much as I tried previous HRTF samples, nice! :-) However, I added a JOSResample and ChannelMapper to be able to play any sound. Also the error messages/crashes that I get when no HRTFs could be loaded for example are not really graceful.

The functionality to load the HRTFs from a directory should be added to the HRTF class maybe, or a separate loader class, then you can have the platform specific code nicely separated from the rest and there's no need for duplicated code.

DethonUSAL:
Thanks! I'll look into the error messages.
Regarding the HRTFs load, I agree a loader class would be nice. A certain HRTF naming scheme would be needed though, so I think I'll keep the KEMAR's one.

include/util/ThreadPool.h
```diff
@@ -0,0 +1,148 @@
...
+	ThreadPool& operator=(const ThreadPool&) = delete;
+public:
+	/**
+	* Creates a new ThreadPool object.
+	* \param count The number of threads of the pool. It must not be 0.
+	*/
+	ThreadPool(unsigned int count) :
+		m_stopFlag(false), m_numThreads(count)
+	{
+		for(unsigned int i = 0; i < count; i++)
```
neXyon: Why is this whole class implemented in the header?

DethonUSAL: Because it has a template method, but, thinking about it, I believe that including a .cpp file in the header could work. I'll try that in a bit.

neXyon: Hmm, no if it's not possible to move any functions to a cpp file, there's no need to do that, but I thought that the functions that are not a template function could be moved into a cpp file, as they are not necessary to be recompiled for every template argument?!

DethonUSAL: I think that is indeed possible. Haven't tested though, I usually like to have all the class implementation in one place.

neXyon: There are many reasons to keep as much code as possible outside of headers:

- compilation time (every cpp including the header needs to compile all of it)
- binary size (the functions are compiled into machine code once for every cpp file and if the linker is not able to optimize that, they are all included in the binary; or it "optimizes" by inlining guaranteeing the unwanted duplication)
- unnecessary recompilation (every time you change something inside the functions that would be in the cpp, all cpps depending on the header get recompiled)
- users of the library not having to worry about the implementation (they just use headers, so they just worry about what is in there)
- the principle of separating interface and implementation (which unfortunately is not easily possible with templates)
- probably there are more reasons

DethonUSAL: You are right, I never really thought about it. Thanks!

demos/binaural.cpp
```diff
@@ -0,0 +1,117 @@
...
+#include "devices/IDevice.h"
+#include "devices/IDeviceFactory.h"
+#include "devices/IHandle.h"
+#include "plugin/PluginManager.h"
+#include "file/File.h"
+#include "respec/ChannelMapper.h"
+#include "respec/JOSResample.h"
+#include "Exception.h"
+
+#include <windows.h>
```
neXyon: Please remove unnecessary headers again. windows.h makes compilation fail on non-windows platforms.

PR #4, opened 2016-01-31
Bug fixing and polish for Audaspace 1.2

DethonUSAL:
Aside from the bug fixing, I also added convolution and binaural sounds to the C and Python APIs.


bindings/C/AUD_HRTF.h
```diff
@@ -0,0 +1,48 @@
...
+extern AUD_API void AUD_HRTF_free(AUD_HRTF* hrtfs);
+
+/**
+* Adds a new impulse response to an HRTF object.
+* \param hrtfs The HRTF object.
+* \param sound A Sound object representing an HRTF.
+* \param azimuth The azimuth angle of the HRTF.
+* \param elevation The elevation angle of the HRTF.
+*/
+extern AUD_API void AUD_HRTF_addImpulseResponse(AUD_HRTF* hrtfs, AUD_Sound* sound, float azimuth, float elevation);
```
neXyon: Why are you adding a Sound instead of an ImpulseResponse? That's a bit confusing, I guess.

DethonUSAL: Hmmm, It's because the HRTF class has no method to add directly an ImpulseResponse object. This was to prevent the user from adding ImpulseResponses that used FFTPlan objects with different size.
I agree it's a bit confusing, so I'll change it and add some documentation about it.

neXyon: That is a valid and very good reason, I recommend to just rename the method to addImpulseResponseFromSound and you're done. :-)

PR #25, opened 2020-04-17
Mixer sample buffer added channels twice

DarkDefender:
Perhaps I'm missing something, but when I looked through the code when trying to hunt down some A/V sync issues, I noticed that channels are added twice here. (Once with `m_specs.channels *` and once more in the `AUD_SAMPLE_SIZE` macro)

This was the only place where this happens where `AUD_SAMPLE_SIZE` is used, so I guess it is a mistake?


neXyon:
Thanks, besides wasting a little bit of memory, this bug shouldn't cause any issues though.

PR #26, opened 2020-04-17
Simplify buffer resize function

DarkDefender:
Also stumbled upon this when looking into A/V sync issues.

The "keep" code branch did the same thing as std::realloc.
(Unless I'm mistaken...)

### DIFF
--- a/bindings/C/AUD_Special.cpp
+++ b/bindings/C/AUD_Special.cpp
@@ -153,7 +153,7 @@ AUD_API float* AUD_readSoundBuffer(const char* filename, float low, float high,
 		do
 		{
 			len = samplerate;
-			buffer.resize((position + len) * sizeof(float), true);
+			buffer.resize((position + len) * sizeof(float));
 			reader->read(len, eos, buffer.getBuffer() + position);
 			position += len;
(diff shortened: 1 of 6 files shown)


neXyon:
Unfortunately, it's not that easy. A simple realloc does not necessarily keep the alignment, that's why that extended bit of code is necessary. In the case where no keeping of the memory is necessary, you could of course do a free + malloc instead of realloc, but a realloc may be faster in many cases (e.g. when the size decreases), while of course risking that it may be slower, when the memory needs to be copied unnecessarily. I would of course prefer a function to check if realloc without copy is possible to decide which method to use, but unfortunately, this doesn't exist afaik.

DarkDefender:
Aha, I see! Thanks for the clarification.

Perhaps you could add a comment in the code with this information?

It tripped up me and and other core blender dev. Because realloc states that it will keep the data from the previous memory adress (if shrinking it will of course lose the data at the end).

So to us the `keep` flag was highly confusing. Because in both cases you do essentially keep the data. Therefore clarifying that you use realloc for simplicity and that it is not equivalent would be good I feel.

PR #27, opened 2020-04-17
Add proof of concept high resolution audio syncing

DarkDefender:
When working with Blender, I've always thought that the `A/V Sync` option for playback was jerky and non smooth.

I took a look at it and it turns out that I was right!
There is some code in place that tries to mask it, but if you remove it, the issue become really apparent.
(Blender patch:)
```
--- a/source/blender/editors/screen/screen_ops.c
+++ b/source/blender/editors/screen/screen_ops.c
@@ -4412,14 +4412,8 @@ static int screen_animation_step(bContext *C, wmOperator *UNUSED(op), const wmEv
     else if ((scene->audio.flag & AUDIO_SYNC) && (sad->flag & ANIMPLAY_FLAG_REVERSE) == false &&
              isfinite(time = BKE_sound_sync_scene(scene_eval))) {
       double newfra = (double)time * FPS;
-
-      /* give some space here to avoid jumps */
-      if (newfra + 0.5 > scene->r.cfra && newfra - 0.5 < scene->r.cfra) {
-        scene->r.cfra++;
-      }
-      else {
-        scene->r.cfra = max_ii(scene->r.cfra, newfra + 0.5);
-      }
+      scene->r.cfra = newfra;
+      scene->r.subframe = fractf(newfra);  // TODO is this needed or nice to have?

 #ifdef PROFILE_AUDIO_SYNCH
       newfra_int = scene->r.cfra;
```

The issue is that, at least for OpenAL and SDL on my computer, the audio position query resolution is not high enough. So when asking for the current time, we will get freeze frames.

This is very apparent with big buffer sizes on the SDL backend.
With OpenAL this is not as noticeable, but it is still there to a lesser extent.
You can tell the difference in smoothness when toggling between `No Sync` and `A/V sync`.

I looked around online to see what the common approach is to solve this, and it seems like the simplest way is to have a timer running so we keep track of the time ourselves and then use the audio sync frames to adjust the drift.

Here is the  results with the SDL backend and a buffer size of 16384:
https://wiki.blender.org/w/images/2/2d/Old.mp4

And with my changes:
https://wiki.blender.org/w/images/f/fd/AV_sync_mk2.mp4

With my changes, I have a really hard time telling `A/V Sync` and `No Sync` apart :)

This is only a quick and dirty test to add higher resolution audio
position reporting to the SDL backend.

I don't expect this to go in as is at all. It is mostly so we can discuss how to implement this in a good way.

One really bad thing I noticed though:
If I have no audio playing on my system at all, the new reads of audio buffers becomes really sporadic and doesn't match the flow of time at all. I guess this might be a optimization of some sorts. But this is not an issue with this patch per se.

### DIFF
--- a/include/devices/SoftwareDevice.h
+++ b/include/devices/SoftwareDevice.h
@@ -31,6 +31,7 @@
 
 #include <list>
 #include <mutex>
+#include <chrono>
 
 AUD_NAMESPACE_BEGIN
 
@@ -60,6 +61,13 @@ protected:
 		SoftwareHandle& operator=(const SoftwareHandle&) = delete;
 
 	public:
+		/// The last time that the buffer position changed
+		std::chrono::time_point<std::chrono::steady_clock> t1;
+		int prev_pos = 0;
+		float offset = 0;
+		float adjust = 0;
+		bool reset_clock = true;
+
 		/// The reader source.
 		std::shared_ptr<IReader> m_reader;
 
--- a/src/devices/SoftwareDevice.cpp
+++ b/src/devices/SoftwareDevice.cpp
@@ -29,6 +29,7 @@
 #include <iostream>
 #include <limits>
 #include <mutex>
+#include <chrono>
 
 AUD_NAMESPACE_BEGIN
 
@@ -272,6 +273,9 @@ bool SoftwareDevice::SoftwareHandle::resume()
 						m_device->playing(m_device->m_playback = true);
 					m_status = STATUS_PLAYING;
 
+					reset_clock = true;
+					printf("reset clock\n");
+
 					return true;
 				}
 			}
@@ -363,11 +367,17 @@ bool SoftwareDevice::SoftwareHandle::seek(float position)
 	if(m_status == STATUS_STOPPED)
 		m_status = STATUS_PAUSED;
 
+	offset = 0;
+	prev_pos = m_reader->getPosition();
+	printf("init position\n");
+
 	return true;
 }
 
 float SoftwareDevice::SoftwareHandle::getPosition()
 {
+	float position;
+
 	if(!m_status)
 		return false;
 
@@ -376,7 +386,55 @@ float SoftwareDevice::SoftwareHandle::getPosition()
 	if(!m_status)
 		return 0.0f;
 
-	float position = m_reader->getPosition() / (float)m_device->m_specs.rate;
+	if(m_status != STATUS_PLAYING) {
+		printf("not playing\n");
+		position = m_reader->getPosition() / (float)m_device->m_specs.rate;
+	} else {
+		using namespace std::chrono;
+
+		if (reset_clock) {
+			reset_clock = false;
+			t1 = steady_clock::now();
+		}
+
+		int new_pos = m_reader->getPosition();
+		steady_clock::time_point t2 = steady_clock::now();
+		duration<double> time_span = duration_cast<duration<double>>(t2 - t1);
+
+		if (prev_pos == new_pos) {
+			offset += time_span.count() + adjust;
+			//printf("offset %f\n", offset);
+		} else {
+			//printf("===== Sync!\n");
+			float time_step = (new_pos - prev_pos) / (float)m_device->m_specs.rate;
+
+			offset = offset - time_step;
+			offset += time_span.count();
+			//printf("time step %f\n", time_step);
+			printf("time diff %f\n", offset);
+			printf("adjust %f\n", adjust);
+
+			if (fabs(offset) > 0.02f) {
+				//TODO the offset error margin and adjust variable should be tied to playback FPS
+				printf("reset!\n");
+				//Try to minimize the offset by slowly drifting towards the minimal error margin.
+				adjust = 0.001f;
+				adjust = std::copysign(adjust,offset);
+				adjust *= -1.0f;
+			} else {
+				adjust = 0;
+			}
+
+			prev_pos = new_pos;
+		}
+
+		t1 = steady_clock::now();
+
+		position = new_pos / (float)m_device->m_specs.rate;
+		position += offset;
+	}
+
+	//printf("position: %f\n", position);
 
 	return position;
 }


neXyon:
This is interesting! Blender has three synchronization methods: No Sync, AV Sync and Frame Dropping. What you are proposing here is a hybrid between the latter two, since Frame Dropping is essentially using a timer. Have you tried if it works accurately enough for you, not incorporating any drift from the audio? If not it might be an idea to add a fourth one, that does what you are proposing here.

I would not always impose it though (by implementing it within the `SoftwareDevice`) since SDL is not the only backend using this class. Since this is more of a Blender issue than an audaspace issue, I'd rather see this problem solved in Blender code, not the audaspace code. Users of the audaspace library would expect the playback time returned to be dependent on the audio only, not an additional timer, of course to the accuracy possible with the actual audio backend.

Therefore, I'll close this here, but please keep answering and if you have something for Blender, feel free to create a patch on https://developer.blender.org/ and add me for review!

DarkDefender:
I do agree that we perhaps shouldn't make this default so that if users want the get the raw resolution time reporting, they can still do so.

I can of course implement this solely on the Blender side of things. However are you interested in perhaps having a raw and an other interpolated getter for time perhaps?
For blender it doesn't really matter where we implement this I guess.
One thing I would like though is to be able to get the raw buffer location int so I do not have to compare floats to see if the time changed.

So what I am now proposing instead is that we lift the interpolation part outside from the individual backends and instead have the interpolation happen at a higher level. (This can of course be inside the blender code if you insist)

I'm not sure what you are proposing for the forth option. If I do not try to account for drift, then this would essentially just a be a glorified "Frame dropping" method.
It would sync at the start of playback but then not actually try to make sure that the audio sync up with the video.

I also looked at the commended sync settings:
https://en.wikipedia.org/wiki/Audio-to-video_synchronization#Recommendations

This is why I have the +/- 20ms error threshold in the code right now. However I realized that depending on the buffer sizes and playback framerate, we could grow/shrink this margin so that we do not over or under correct.

In my tests, I have been able to keep a 60fps playback in sync with very nice margins (10-20ms).
However I do see that when at least using the SDL backend, it will have to adjust it self for drift from time to time even after it has homed in. (Perhaps this also answers your previous question?)

neXyon:
> Have you tried if it works accurately enough for you, not incorporating any drift from the audio?

With that I meant, whether you tried if Frame Dropping works accurately enough for you? The fourth/new should of course have the drift compensation. That's what I mean with the hybrid and the whole point of it. I would be curious how big the drift gets over time! Can you provide some numbers for your system?

> I can of course implement this solely on the Blender side of things. However are you interested in perhaps having a raw and an other interpolated getter for time perhaps?

Yes, please! If you like you could however also add another `ISynchronizer` implementation that builds on top of another `ISynchronizer` to do this.

> One thing I would like though is to be able to get the raw buffer location int so I do not have to compare floats to see if the time changed.

I'm not really happy with a float here either, but I'm even less happy about having two versions of basically the same functionality. It would be possible to change the interface, but that would break compatibility. I think the least intrusive change that also keeps the API as clean as possible is changing the datatype for `seek` and `getPosition` in `IHandle` and `ISynchronizer` from `float` to `double`. This would make it accurate enough for hundreds of years rather than a few minutes. Or do you have any reasons why a double would be bad?

DarkDefender:
For me it is mostly that the an integer is the actual representation of the buffer position.
And in this case that is ideally what I want to work with to see if the reported position has changed since the last frame.

If that is not the case, then we will have to interpolate to make playback smooth.
If the position has changed then it means that we can try to calculate the drift and compensate for it.

I guess using floats or doubles here wouldn't be that bad, it simply irks me a bit that we have something exact and then goes to something non exact. (Though I don't think it will be a problem in practice).

But why would introducing an alternative getter break compatibility?  If we don't touch the things already in place, this wouldn't be a problem, no?

DarkDefender:
About the drift numbers:

If I start playing back a 10000 frame sequence at 60fps (so around 3 minutes of playback), the highest initial time difference between the reported time and the calculated time at the begining for playback was:
`time diff -0.064666`

Then after playing back around 10000 frames, the highest was:
`Time diff -0.116938`

The the average time diff increased during this time steadily so that implies that there is an actual drift and not just that the extremes got higher.

So I would estimate that during this test we drifted around 60ms off sync.

If I then plug back in my time drift compensation code, I can see this detects drift and adjusts for it every now and again. With this code the average offset is within +/- 20ms. Most sync frames are within 20ms, but the outliers that triggers the adjustment  seems to be within the 20-25ms range.

neXyon:
> For me it is mostly that the an integer is the actual representation of the buffer position.
> And in this case that is ideally what I want to work with to see if the reported position has changed since the last frame.
> 
> If that is not the case, then we will have to interpolate to make playback smooth.
> If the position has changed then it means that we can try to calculate the drift and compensate for it.
> 
> I guess using floats or doubles here wouldn't be that bad, it simply irks me a bit that we have something exact and then go to something non exact. (Though I don't think it will be a problem in practice).
> 
> But why would introducing an alternative getter break compatibility? If we don't touch the things already in place, this wouldn't be a problem, no?

neXyon:
> I guess using floats or doubles here wouldn't be that bad, it simply irks me a bit that we have something exact and then go to something non exact. (Though I don't think it will be a problem in practice).
> 
> But why would introducing an alternative getter break compatibility? If we don't touch the things already in place, this wouldn't be a problem, no?

I'm suggesting doubles because at a sampling rate of 48 kHz, you'd be sample accurate for thousands of years and I don't think anyone will work with time spans that high. In that sense I wouldn't say that they are not exact. I agree in the case of floats though, since there it's only accurate for a few minutes.

I didn't say the alternative would break compatibility. I don't like having two methods that provide the same functionality, since that would be bad API design in my opinion. Avoiding two methods and instead switching to samples instead of time as the only one would break compatibility.

> So I would estimate that during this test we drifted around 60ms off sync.

After less than 3 minutes, that's considerable, so it definitely would make sense to add this synchronization option.

DarkDefender:
> I didn't say the alternative would break compatibility. I don't like having two methods that provide the same functionality, since that would be bad API design in my opinion. Avoiding two methods and instead switching to samples instead of time as the only one would break compatibility.

To me it doesn't necessarily provide the same functionality.
The current function provides the estimated time (in float of double).

The new one would just flat out return the current buffer position. So it would basically be just:
```
int getBufferPosition() {
  return m_reader->getPosition()
}
```

Because IIRC, we can't access `m_reader` from outside.

But if you just want to go doubles instead, we can simply do that.

neXyon:
There is no misunderstanding. For the high level API of the library I simply prefer to work with times and not have the users worry about sample positions as much as possible. It is clear what time is, but what exactly is a sample position? You call it buffer position which is already a wrong term, since in most cases, there is no buffer behind the handle/reader. Is it the sample count? Is a sample a single value or multiple for the other channels? Do you then have to also divide by the number of channels? A user of the API doesn't have to think about these problems if it simply returns the time.

I have incorporated the changes to double for more than enough precision. They are also already ported to blender's master branch for you to play  around with :)

DarkDefender:
BTW, do you want me to open up a different pull request for the new implementation based on `ISynchronizer`?

Or should we continue here when that happens?

PR #28, opened 2020-05-08
Add proof of concept high resolution audio syncing V2

DarkDefender:
Continuation of #27

I replaced the default synchronizer in this change, but I guess I should just create a `InterpolatedSyncronizer` and have the audio backends be able to choose between this and the `DefaultSynchronizer`?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -30,6 +30,7 @@ set(AUDASPACE_LONG_VERSION ${AUDASPACE_VERSION}.0)
 
 set(SRC
 	src/devices/DefaultSynchronizer.cpp
+	src/devices/InterpolatedSynchronizer.cpp
 	src/devices/DeviceManager.cpp
 	src/devices/NULLDevice.cpp
 	src/devices/ReadDevice.cpp
@@ -129,7 +130,7 @@ set(PRIVATE_HDR
 )
 
 set(PUBLIC_HDR
-	include/devices/DefaultSynchronizer.h
+	include/devices/InterpolatedSynchronizer.h
 	include/devices/DeviceManager.h
 	include/devices/I3DDevice.h
 	include/devices/I3DHandle.h
--- a/bindings/C/AUD_Device.cpp
+++ b/bindings/C/AUD_Device.cpp
@@ -305,6 +305,11 @@ AUD_API double AUD_getSynchronizerPosition(AUD_Handle* handle)
 	return (*reinterpret_cast<std::shared_ptr<IHandle>*>(handle))->getPosition();
 }
 
+AUD_API void AUD_setInterpolatedSynchronizer(bool interpolated)
+{
+	DeviceManager::getDevice()->createSynchronizer(interpolated);
+}
+
 AUD_API void AUD_playSynchronizer()
 {
 	auto synchronizer = DeviceManager::getDevice()->getSynchronizer();
--- a/bindings/C/AUD_Device.h
+++ b/bindings/C/AUD_Device.h
@@ -231,6 +231,15 @@ extern AUD_API void AUD_seekSynchronizer(AUD_Handle* handle, double time);
 extern AUD_API double AUD_getSynchronizerPosition(AUD_Handle* handle);
 
 /**
+ * Sets the syncronizer type to interpolated if 'true' is passed.
+ * Per default the syncronizer is not interpolated.
+ * Will only have an effect if the device supports or might benefit from
+ * interpolation.
+ * \param interpolated Set the syncronizer to be interpolated or not.
+ */
+extern AUD_API void AUD_setInterpolatedSynchronizer(bool interpolated);
+
+/**
  * Starts the playback of jack transport if possible.
  */
 extern AUD_API void AUD_playSynchronizer();
--- a/include/devices/IDevice.h
+++ b/include/devices/IDevice.h
@@ -24,6 +24,7 @@
 
 #include "respec/Specification.h"
 #include "util/ILockable.h"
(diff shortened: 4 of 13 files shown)


neXyon:
Sorry, super busy these days, will get better next week.

If you want to implement this within audaspace and not within blender, then yes, this should be a separate class, definitely. It should be possible to plug this on top of another synchronizer, so that you can use it independently of the device and the device doesn't have to know about it.

DarkDefender:
I'm not quite sure how you want the API to look like.
Could you help a bit with this when you have time?

neXyon:
Sure! First, you should put your code into its own class that implements ISynchronizer and that also has an ISynchronizer it builds on (e.g. by getting another ISynchronizer pointer in the constructor).
Then, since you want to use this from the blender source code, you'll have to make this class accessible through the C API. Currently, the C API only has functions to work with the synchronizer of the device that is currently set within the DeviceManager. I see two options there: you either add functionality to set another synchronizer on top using either the DeviceManager or a new SynchronizerManager which then manages the instance of the ISynchronizer, or you create a straightforward C API for Synchronizers and then the instance has to be managed from within blender. The latter option might be the more difficult one though.

DarkDefender:
Updated so this can be set at run time via C bindings

neXyon:
Ok, then let me detail a little what I think of the `DeviceManager` route. I'll give you some steps of what would be to do there:

- The following changes in your `InterpolatedSynchronizer` class are required to be able to "stack" synchronizers on top of each other as discussed above already:
  - You add a pointer to another synchronizer as a member `m_synchronizer`, that is set based on a parameter in the (newly added) constructor.
  - The interface methods that currently call the functions of the handle (`handle->seek(time)` in `seek` and `handle->getPosition()` in `getPosition` call the respective methods of that synchronizer instead of the handle directly.
  - The remaining interface methods (`play`, `stop`, `setSyncCallback` and `isPlaying`) also forward the call to `m_synchronizer`.
- Revert all changes to device classes (`IDevice`, `NULLDevice`, `SoftwareDevice`, `JackDevice` and `OpenALDevice`), since the devices don't care if you want to use the `InterpolatedSynchronizer` on top of them and we thus don't want to change their API, as you do by changing `IDevice`.
- Change the API of the `DeviceManager` to also manage a synchronizer instance as I meant above:
  - Add a new `static std::shared_ptr<ISynchronizer> m_synchronizer;` member to the `DeviceManager` class.
  - Add three static methods do set, get and clear it (`setSynchronizer`, `getSynchronizer` and `releaseSynchronizer`) as follows:
  - The `getSynchronizer()` method you should return an `ISynchronizer*` and not the `shared_ptr` and you also have to check the shared_ptr member first. If it's cleared/not set (i.e. the `shared_ptr` is a `nullptr`), you should return the device's synchronizer (`m_device->getSynchronizer()').
  - The `setSynchronizer` method expects a `shared_ptr` and simply sets `m_synchronizer` to it.
  - The `releaseSynchronizer` method simply releases `m_synchronizer` (`m_synchronizer.reset()`).
- Finally, you can add (change) functions to the C API to be able to use your `InterpolatedSynchronizer`, changing `AUD_Device.cpp/h`:
  - Remove the parameter of `AUD_setInterpolatedSynchronizer` and in the function's body create a new instance of a `InterpolatedSynchronizer` in a `shared_ptr` passing in `DeviceManager::getSynchronizer()` as constructor parameter and then call `DeviceManager::setSynchronizer()` with it.
  - Add a function `AUD_releaseSynchronizer()` that simply calls `DeviceManager::releaseSynchronizer()`.

That should be it, if I didn't miss something :sweat_smile:

DarkDefender:
Sorry for not being able to merge in your changes yet.

Before that though I did some more testing and have some questions as I think we might need to iterate a bit more on the core functionality.

I noticed that when starting to play, there is usually some lag before the audio starts. This is usually not an issue when playing back longer animations as the new sync code will make the animation catch up (or slow down) so it matches the audio.

However when playing back smaller loops or when trying to start playback at a critical point (like a audio cue or lip sync for example), this becomes quite annoying.

It seems like even when looping the audio is not prefetched (because we probably lack any way to communicate this...), so on each loop the sync is a bit different at the start.

This is really bad as it makes it kinda unusable just at the start of playback.

So in this sense the older method was better as the playback would remain at the start point until the audio system reported that it had advanced.
So I'm asking you if there perhaps is anyway we could make this nicer?

For just "cold starts" I'm guessing that SDL and OpenAL doesn't support reporting back when it actually starts playback?

If this is the case, then we could "warm up" the audio cache a bit by starting to play back some silence in the background and start playback. During my testing these cold starts have a lag time of around 50-300ms on my computer when starting playback.

But this is a bit tricky too as we probably can't tell exactly when the silent clip has finished playing... Hmm...

For loops though, I guess we could perhaps add some more communication between Blender and audaspace so we prefetch audio into the buffer before it loops around. This way we could hopefully eliminate the jerkiness when jumping from end to start.

Now this is just my initial thoughts. They might be horrible ideas or just flat out wrong. So this is why I am asking you as the audio guru for help.

neXyon:
Hmm, I'm honestly not a big fan of adding an artificial delay here. Maybe it's better to give a higher weight on the audio reported playback position when playback starts or is seeked, i.e. when looping and then gradually increase the correction of the timer?

DarkDefender:
Just so I don't leave you hanging:

I will probably not have time to work on this for a while now.
So it will take a while until I can continue with this.

neXyon:
Ok, no worries. I'll close this merge request for now. Feel free to open it again as soon, as you have time to work on it again.

PR #39, opened 2024-05-14
Update FFmpeg code to avoid direct assignation of ch_layout

eclipseo:
> Copying an AVChannelLayout via assigning is forbidden, av_channel_layout_copy()
> must be used instead (and its return value should be checked)

so we make use of av_channel_layout_copy()


neXyon:
Thanks for the fix!

PR #42, opened 2024-09-04
Audaspace: Update PulseAudio sync functions to be more accurate

DarkDefender:
As requested in https://projects.blender.org/blender/blender/pulls/126091, here is the pulseaudio changes.

### DIFF
--- a/plugins/pulseaudio/PulseAudioDevice.cpp
+++ b/plugins/pulseaudio/PulseAudioDevice.cpp
@@ -15,28 +15,39 @@
  ******************************************************************************/
 
 #include "PulseAudioDevice.h"
-#include "PulseAudioLibrary.h"
-#include "devices/DeviceManager.h"
-#include "devices/IDeviceFactory.h"
+
 #include "Exception.h"
 #include "IReader.h"
+#include "PulseAudioLibrary.h"
+
+#include "devices/DeviceManager.h"
+#include "devices/IDeviceFactory.h"
 
 AUD_NAMESPACE_BEGIN
 
-PulseAudioDevice::PulseAudioSynchronizer::PulseAudioSynchronizer(PulseAudioDevice *device) :
-	m_device(device)
+PulseAudioDevice::PulseAudioSynchronizer::PulseAudioSynchronizer(PulseAudioDevice* device) : m_device(device)
 {
+}
+
+void PulseAudioDevice::PulseAudioSynchronizer::update_time_start()
+{
+	AUD_pa_stream_get_time(m_device->m_stream, &m_time_start);
+	m_seek_pos = m_timeline_pos;
+}
+
+void PulseAudioDevice::PulseAudioSynchronizer::seek(std::shared_ptr<IHandle> handle, double time)
+{
+	AUD_pa_stream_get_time(m_device->m_stream, &m_time_start);
+	m_seek_pos = m_timeline_pos = time;
+	handle->seek(time);
 }
 
 double PulseAudioDevice::PulseAudioSynchronizer::getPosition(std::shared_ptr<IHandle> handle)
 {
-	pa_usec_t latency;
-	int negative;
-	AUD_pa_stream_get_latency(m_device->m_stream, &latency, &negative);
-
-	double delay = m_device->m_ring_buffer.getReadSize() / (AUD_SAMPLE_SIZE(m_device->m_specs) * m_device->m_specs.rate) + latency * 1.0e-6;
-
-	return handle->getPosition() - delay;
+	pa_usec_t time;
+	AUD_pa_stream_get_time(m_device->m_stream, &time);
+	m_timeline_pos = (time - m_time_start) * 1.0e-6 + m_seek_pos;
+	return m_timeline_pos;
 }
 
 void PulseAudioDevice::updateRingBuffer()
@@ -85,18 +96,18 @@ void PulseAudioDevice::updateRingBuffer()
 	}
 }
 
-void PulseAudioDevice::PulseAudio_state_callback(pa_context *context, void *data)
+void PulseAudioDevice::PulseAudio_state_callback(pa_context* context, void* data)
 {
-	PulseAudioDevice* device = (PulseAudioDevice*)data;
+	PulseAudioDevice* device = (PulseAudioDevice*) data;
 
 	device->m_state = AUD_pa_context_get_state(context);
 
 	AUD_pa_threaded_mainloop_signal(device->m_mainloop, 0);
 }
 
-void PulseAudioDevice::PulseAudio_request(pa_stream *stream, size_t total_bytes, void *data)
+void PulseAudioDevice::PulseAudio_request(pa_stream* stream, size_t total_bytes, void* data)
 {
-	PulseAudioDevice* device = (PulseAudioDevice*)data;
+	PulseAudioDevice* device = (PulseAudioDevice*) data;
 
 	data_t* buffer;
 
@@ -139,6 +150,7 @@ void PulseAudioDevice::playing(bool playing)
 	{
 		AUD_pa_threaded_mainloop_lock(m_mainloop);
 		AUD_pa_stream_cork(m_stream, 0, nullptr, nullptr);
+		m_synchronizer.update_time_start();
 		AUD_pa_threaded_mainloop_unlock(m_mainloop);
 	}
 }
--- a/plugins/pulseaudio/PulseAudioDevice.cpp
+++ b/plugins/pulseaudio/PulseAudioDevice.cpp
@@ -29,8 +29,9 @@ PulseAudioDevice::PulseAudioSynchronizer::PulseAudioSynchronizer(PulseAudioDevic
 {
 }
 
-void PulseAudioDevice::PulseAudioSynchronizer::update_time_start()
+void PulseAudioDevice::PulseAudioSynchronizer::play()
 {
+	/* Make sure that our start time is up to date. */
 	AUD_pa_stream_get_time(m_device->m_stream, &m_time_start);
 	m_seek_pos = m_timeline_pos;
 }
@@ -150,7 +151,6 @@ void PulseAudioDevice::playing(bool playing)
 	{
 		AUD_pa_threaded_mainloop_lock(m_mainloop);
 		AUD_pa_stream_cork(m_stream, 0, nullptr, nullptr);
-		m_synchronizer.update_time_start();
 		AUD_pa_threaded_mainloop_unlock(m_mainloop);
 	}
 }
--- a/plugins/pulseaudio/PulseAudioDevice.cpp
+++ b/plugins/pulseaudio/PulseAudioDevice.cpp
@@ -33,22 +33,33 @@ void PulseAudioDevice::PulseAudioSynchronizer::play()
 {
 	/* Make sure that our start time is up to date. */
 	AUD_pa_stream_get_time(m_device->m_stream, &m_time_start);
-	m_seek_pos = m_timeline_pos;
+	m_playing = true;
+}
+
+void PulseAudioDevice::PulseAudioSynchronizer::stop()
+{
+	std::shared_ptr<IHandle> dummy_handle;
+	m_seek_pos = getPosition(dummy_handle);
+	m_playing = false;
 }
 
 void PulseAudioDevice::PulseAudioSynchronizer::seek(std::shared_ptr<IHandle> handle, double time)
 {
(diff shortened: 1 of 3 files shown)


plugins/pulseaudio/PulseAudioDevice.h
```diff
@@ -45,10 +45,15 @@ class AUD_PLUGIN_API PulseAudioDevice : public SoftwareDevice
...
 	{
 		PulseAudioDevice* m_device;
+		pa_usec_t m_time_start = 0;
+		double m_seek_pos = 0.0f;
+		double m_timeline_pos = 0.0f;
 
 	public:
 		PulseAudioSynchronizer(PulseAudioDevice* device);
 
+		void update_time_start();
```
neXyon, line 55: Please follow the camelCase naming scheme that audaspace uses for methods. Since you should implement `play()`, this would be renamed anyway.

plugins/pulseaudio/PulseAudioDevice.cpp
```diff
@@ -15,28 +15,39 @@
...
+PulseAudioDevice::PulseAudioSynchronizer::PulseAudioSynchronizer(PulseAudioDevice* device) : m_device(device)
 {
 }
 
-double PulseAudioDevice::PulseAudioSynchronizer::getPosition(std::shared_ptr<IHandle> handle)
+void PulseAudioDevice::PulseAudioSynchronizer::update_time_start()
 {
-	pa_usec_t latency;
-	int negative;
-	AUD_pa_stream_get_latency(m_device->m_stream, &latency, &negative);
```
neXyon, line 35: When you seek during playback you should still consider the latency/buffer size since the actual playback of the seeked handle will only happen once the already filled buffers are played back.

plugins/pulseaudio/PulseAudioDevice.cpp
```diff
@@ -139,6 +150,7 @@ void PulseAudioDevice::playing(bool playing)
 	{
 		AUD_pa_threaded_mainloop_lock(m_mainloop);
 		AUD_pa_stream_cork(m_stream, 0, nullptr, nullptr);
+		m_synchronizer.update_time_start();
```
neXyon, line 153: This makes the synchronizer timing depended on when you play back sounds. That's not right. You should implement `ISynchronizer::play()` and set the start time there.

plugins/pulseaudio/PulseAudioDevice.h
```diff
@@ -45,10 +45,15 @@ class AUD_PLUGIN_API PulseAudioDevice : public SoftwareDevice
 	class PulseAudioSynchronizer : public DefaultSynchronizer
 	{
 		PulseAudioDevice* m_device;
+		pa_usec_t m_time_start = 0;
+		double m_seek_pos = 0.0f;
+		double m_timeline_pos = 0.0f;
```
neXyon, line 50: A single`pa_usec_t m_start_time = 0` should be enough. When you seek, you can simply compute it by getting the current time and subtracting the seek time (and consider latency as mentioned).

neXyon (changes requested):

plugins/pulseaudio/PulseAudioDevice.cpp
DarkDefender: I'm a bit unsure what you mean with this.

If you seek during playback and queue up a pause command immediately after, the reported position would be incorrect.

Same thing if you query the position directly after you seek. Lets say you seek to 0 and then the added latency would then make it report a negative seek position. This is exactly what this patch fixes as previously we needed special hacks in Blender to make sure that the actual reported position matched to what we actually were seeking to.

As this only happened in the pulseaudio backend, I don't think adding in latency to the reported position is the correct solution. Perhaps I'm misunderstanding you though?

plugins/pulseaudio/PulseAudioDevice.h
DarkDefender: I don't think that will work as the time counter from pulseaudio will continue to count up with pause (IIRC because we don't always cork the stream when paused).

Maybe I should rework things more to see if we can cut down on these variables?

plugins/pulseaudio/PulseAudioDevice.cpp
neXyon: The old code computed the position as follows:

Get the position of the handle and then we subtract the following two values as the handle might not have been played back that far:

- Parts of what has been played back by the handle might be in the ring buffer waiting to be forwarded to pulse, so it hasn't been played yet.
- Pulse also has a latency which we can think of a buffer that is waiting to be played.

Of course, you're right that if we just seeked or if playback just started for a handle this will result in a time before the seek or a negative number if playback just started.

It is correct though in the sense that it tells you when it will start playing or start playing the position you seeked to. In that sense I don't consider this a bug, it's rather the most precise seeking in comparison to other backends which don't provide this latency information. It's the other backends that are less precise.

I understand that the timeline jumping back a bit in Blender is bad. This might happen of course, depending on the buffer size, the latency and also how the reported audio time is rounded to the blender playhead. Maybe a solution in Blender would be to seek to the middle of a frame, e.g., if rendering runs at 100 FPS and we seek to the second frame, we should seek to "frame" 2.5, i.e., t = 25 ms instead of t = 20 ms.

plugins/pulseaudio/PulseAudioDevice.h
neXyon: You're right probably. Let's compare to above: pa_stream_get_time does consider latency, so you don't have to subtract it anymore. However it does not consider the content of the ring buffer, so to be really accurate you would need to update the timing again with the amount of samples stored in the ring buffer at the time of seeking, which again gives you negative positions right after starting playback/negative offsets after seeking and thus the same behavior. Have you actually tried simply using the default synchronizer, i.e., directly returning the playback handle's position and find out which one of theses two is more accurate? Might actually be the same pretty much, since I would assume the latency would be shorter to the extent of insignificance in comparison to the ring buffer size. You could also try the original method with just adapting with the ring buffer size, which would give you almost the same result as your current implementation if you enabled `PA_STREAM_NOT_MONOTONIC` and have a bad delay prediction.

DarkDefender: > Have you actually tried simply using the default synchronizer, i.e., directly returning the playback handle's position and find out which one of theses two is more accurate?

I have, yes. It doesn't make too much of a difference on low fps counts (24 fps and below). But as you go higher (30, 60, 120 fps) there are jumps/skips in the playback because the position is not updated frequently enough. By using the reported time from `pa_stream_get_time` I get less jumps/skips.
(The pipewire backend is even more accurate in this regard as well.)

> However it does not consider the content of the ring buffer

Hmm, isn't the ring buffer more of a prebuffer?
IE in theory we could fill up the ring buffer with audio data lasting a minute or so because we already know what we are going to playback from the time line.

It should actually reduce the latency because we have already prebuffered the audio data that we are going to send to pulseaudio. So if the ring buffer is already filled, the latency introduced by feeding it to pulseaudio should be minimal.

Or am I getting this wrong?

Of course when seeking or looping, we might need to wait for the ring buffer to clear out and get filled up with new data. However in these cases, we should probably stop counting up the clock until we have filled the ringbuffer and are ready to feed pulseaudio.

For looping sections, I've seen other programs making sure that their audio backend knows about the start/end times so it can pre buffer data from the start when it is approaching the end.

neXyon: I thought a bit about all that. We could actually do this for all backends. By default (if the backend doesn't provide such timing functions), we could simple use the samples played and count from there. That would allow us to remove the handle from the `getPosition` and `seek` methods. We would store the sync callback and call it directly when seek is called (locking the device in the meantime). Of course this can only be as accurate as the buffer size used for the callback, but that should be more precise than the ring buffer which is usually 3x the size.

You don't want to make the ring buffer unnecessarily big. Sure, if it's just the scene playing it would work, but again, the UI could also want to play back sounds in some cases, maybe in plugins and that would then have a 1 minute delay, which is of course unacceptable.

As for the latency that is introduced by the ring buffer, we could simply store the ring buffer read size when play or seek is called and keep the reported position at the start/seek time until we can reduce that to 0. I guess this is what you mean by stop counting up the clock in that case, right?

As for looping sections: this could be done already. In Blender you could put a limiter and a looper on top of the sequencer when you create the playback handle. That would ensure that it loops very precisely. The only thing you have to do then is modulo the reported positions, since even when looping the reported position will increase monotonically. I.e., if you have 250 frames and it reports frame 254, you need to go to frame 254 % 250 = 4.

In that sense, I'm fine with this patch, if that sounds good to you? I would just ask you to do a little clean up and we can merge this. I can implement the changes to the interface and default synchronizer as described at the beginning.

DarkDefender:
I just wanted to give you a heads up that I'm currently sick with covid. (So things might go a bit slower)
I guessing that it might be because I'm a bit brain dead right now, but I'm having a hard time to grasp exactly what changes to the interface you are proposing to the interface and synchronizer.

I do think that having some interpolation even in the underlying API doesn't have it is good. One of the issues we had on the Blender side of things was the people switched off from using A/V sync as the playback wasn't as smooth as without it.
IIRC I implemented something in the SDL backend to smooth out the position reporting.

It is also the reason I've been working on the other audio backends to make it as good as regular V-sync.

DarkDefender:
> You don't want to make the ring buffer unnecessarily big. Sure, if it's just the scene playing it would work, but again, the UI could also want to play back sounds in some cases, maybe in plugins and that would then have a 1 minute delay, which is of course unacceptable.

Would we want some kind of secondary API for these kinds of "play sound NOW" cases?
It feels a bit weird to me how things are now intertwined in that the audio device has to be in a "playing == true" state for these kinds of impromptu sounds to be played back.

For example. Lets say that Blender did have UI notification sounds and I have queued up sounds on the timeline in the VSE off Blender but it is paused. If a notification sound wanted to playback, and I simply switch on the device to play the notification sound (`playing(true)`), it would also start playing back the audio timeline.

Again, perhaps I'm just too braindead and missing something obvious.

neXyon:
> I just wanted to give you a heads up that I'm currently sick with covid. (So things might go a bit slower) I guessing that it might be because I'm a bit brain dead right now, but I'm having a hard time to grasp exactly what changes to the interface you are proposing to the interface and synchronizer.

No worries, I also take long to respond sometimes, when I'm busy with other stuff. Get well soon!

> I do think that having some interpolation even in the underlying API doesn't have it is good. One of the issues we had on the Blender side of things was the people switched off from using A/V sync as the playback wasn't as smooth as without it. IIRC I implemented something in the SDL backend to smooth out the position reporting.
> 
> It is also the reason I've been working on the other audio backends to make it as good as regular V-sync.

Interpolation is another thing that could be added to smooth it out, but you have to be careful there to not cause jumps again, i.e. let the position only increase monotonically.

neXyon:
> I think would we want some kind of secondary API for these kinds of "play sound NOW" cases? It feels a bit weird to me how things are now intertwined in that the audio device has to be in a "playing == true" state for these kinds of impromptu sounds to be played back.
> 
> For example. Lets say that Blender did have UI notification sounds and I have also queued sounds on the timeline in the VSE of Blender but it is paused. If a notification sound wanted to playback, and I simply switch on the device to play the notification sound (`playing(true)`), it would also start playing back the audio timeline.
> 
> Again, perhaps I'm just too braindead and missing something obvious.

I think you are mixing up two things here. First, there is the playback state of the device. This is mostly internal to the device to be able to shutdown/pause/sleep the backend whenever there is actually nothing playing. This can of course not be changed.

The other playing state is that of the synchronizer and that one is independent of the playing state of the device. That is the one that should be used for synchronization in Blender with the timeline. It's of course not used in the case of notification sounds and the like.

DarkDefender:
I think I understand now. Thank you so much for you explanation and patience!

I realized that my current implemention would break if the end user didn't regularly call `getPosition`. (As the `m_timeline_pos` variable only got updated when calling `getPosition`.

I've updated the code to get rid of `m_timeline_pos` and as a bonus is should now accurately report the position even if the user didn't call `getPosition` during playback. (Of course this doesn't take into account the ring buffer latency.

Note that I only test it briefly as I just wanted to check if you are ok with the changes.

Regarding the latency and general API discussions:

> As for the latency that is introduced by the ring buffer, we could simply store the ring buffer read size when play or seek is called and keep the reported position at the start/seek time until we can reduce that to 0. I guess this is what you mean by stop counting up the clock in that case, right?

Yes, that is what I had in mind.

However I feel it might be best to do all of this later, right?

PR #43, opened 2024-09-04
Pipewire support 

DarkDefender:
As requested in https://projects.blender.org/blender/blender/pulls/126091 here is the Pipewire plugin changes.

This is still WIP though as I want to discuss the `CMakeLists.txt` file.
I noticed that the file was not the same as in the Blender project.
Notably `AUDASPACE_STANDALONE` variable is only in the Blender repo.
(There are of course other things that have not been synced/updated in there, but it is the major difference I think)

Perhaps it would be best to try to keep these files in sync even if the extra variables are only relevant if you want to integrate Audaspace as it is done in Blender?

Of course I can just manually port over the CMakeLists changes. I'm just asking as I think it might be easier to keep audaspace in Blender and this repo in sync if there are as little differences as possible.

If you think it is a good idea, I'll open up an other PR to sync the `CMakeLists.txt` files.


neXyon (changes requested):
I just had a quick look at the synchronizer now, since I just reviewed your patch for pulseaudio and this seems to work pretty much the same. Since pipewire also supports jack, I would have assumed that the API also provides the same functionality for a global clock, so this should work like the Jack synchronizer, no? That would allow to synchronize to other applications using jack/pipewire which is the whole point of the synchronizer class in the first place.

Will look at the rest when I find more time!

plugins/pipewire/PipeWireDevice.h
```diff
@@ -0,0 +1,104 @@
...
+	pw_stream* m_stream;
+	pw_thread_loop* m_thread;
+	std::unique_ptr<pw_stream_events> m_events;
+
+	AUD_LOCAL static void handle_state_changed(void* device_ptr, enum pw_stream_state old, enum pw_stream_state state, const char* error);
+	/**
+	 * Mixes the next bytes into the buffer.
+	 * \param data The PipeWire device.
+	 */
+	AUD_LOCAL static void mix_audio_buffer(void* device_ptr);
```
neXyon, line 73: Please stick with audaspace's camelCase naming convention for methods when possible.

plugins/pipewire/PipeWireDevice.cpp
```diff
@@ -0,0 +1,300 @@
...
+	{
+		/* When we activate/unpause the stream, we need to make sure that
+		 * we update the refrence tick number as the ticks we get from PipeWire
+		 * might have continued to count up after we paused the stream.
+		 */
+		device->m_synchronizer.update_tick_start();
+	}
+}
+
+void PipeWireDevice::mix_audio_buffer(void* device_ptr)
```
neXyon, line 80: Like other devices which have such a callback, it would be much better to have a lightweight and lockless callback that returns as quickly as possible with mixing actually happening in a separate thread. Basically like Jack and pulseaudio are doing it.

plugins/pipewire/PipeWireDevice.cpp
```diff
@@ -0,0 +1,300 @@
...
+		AUD_pw_thread_loop_destroy(m_thread);
+		AUD_THROW(DeviceException, "Could not create PipeWire stream.");
+	}
+
+	spa_audio_info_raw info{};
+	info.channels = m_specs.channels;
+	info.format = format;
+	info.rate = m_specs.rate;
+
+	const spa_pod *params[1];
```
neXyon, line 207: What's the reasoning behind having an array here? A single pointer should suffice (using `&params` below instead of params).

neXyon (changes requested):
Thanks for this contribution. Most of my comments are minor. I didn't want to have all the `AUDASPACE_STANDALONE` ifs in the upstream CMakeLists.txt, but you're right that it's easier to maintain a single version of this file, so this changes is fine (with the commented fixes implemented).

Two major things I would like to see addressed are the previous comment about the synchronizer and a separate mixing thread with a lightweight callback for better real-time support. After all Pipewire raises the claim to be as pro-audio as Jack is, so we should support everything Jack does if possible.

DarkDefender:
> Since pipewire also supports jack, I would have assumed that the API also provides the same functionality for a global clock, so this should work like the Jack synchronizer, no? That would allow to synchronize to other applications using jack/pipewire which is the whole point of the synchronizer class in the first place.
> 

As far as I understand there is not global clock in pipewire. They do implement their own Jack server, so if you want a global clock, you would use the Jack API.

plugins/pipewire/PipeWireDevice.cpp
DarkDefender: This might explain why I got higher reported latency in `pw-top` with this backend than if I were using the Jack or the Pulseaudio backend... :sweat_smile: 

Thanks for pointing this out!

CMakeLists.txt
```diff
@@ -1011,7 +1132,9 @@ if(WITH_C)
 		)
 	endif()
 
-	install(FILES ${C_HDR} DESTINATION include/audaspace)
+	if(AUDASPACE_STANDALONE)
+		install(FILES ${C_HDR} DESTINATION include/audaspace)
+	endif()
```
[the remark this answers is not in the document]
DarkDefender: I've added a new commit with the requested changes.
I'm not sure that I got this comment correct though.

I think you are talking about the `WITH_VERSIONED_PLUGINS` changes?
Or are you talking about something specific to this exact line?

plugins/pipewire/PipeWireDevice.cpp
DarkDefender: I simply copy pasted some of the Pipewire code examples for this part, so no reasoning really.
Will change this to not be an array as you suggested.

CMakeLists.txt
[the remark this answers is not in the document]
neXyon: No clue why this comment is here, but yeah, the `WITH_VERSIONED_PLUGINS` was a change I wanted, thanks!

neXyon:
> > Since pipewire also supports jack, I would have assumed that the API also provides the same functionality for a global clock, so this should work like the Jack synchronizer, no? That would allow to synchronize to other applications using jack/pipewire which is the whole point of the synchronizer class in the first place.
> 
> As far as I understand there is no global clock/sync in pipewire. They do implement their own Jack server, so if you want a global clock, you would use the Jack API. (Please correct me if I'm wrong though as I'm of course not 100% certain about this, but I could find anything when looking at the API)

It should definitely be possible, see the second point here: https://www.phoronix.com/news/PipeWire-1.2-Released

> The concept of node.sync-group was added. This groups nodes with overlapping sync-group together when one of them sets the node.sync = true. This is now used to make sure all nodes are scheduled together when JACK transport is started so that they all see the same transport time.

I just don't know how to do this exactly. One suggestion I heard was looking at this example: https://gitlab.freedesktop.org/pvir/pipewire/-/blob/example-midi-1/src/examples/midi-src.c#L169

That's not enough obviously though since that just synchronizes to the transport position and doesn't set it (seek).

DarkDefender:
I've updated the code to use ringbuffers to pre-buffer the audio data. The latency did indeed improve by quite a bit! (~300us -> ~20us)

For the audio syncing, I looked into it but I don't think we should implement it. At least not in this PR. Here are my reasons:

1. The midi example you found is not in the main repo, so I'm guessing that was the WIP branch and they removed the sync example for the final version. (So there is no official example files showing how this should work)

2. After looking into the pipewire source code, it seems like we have to setup a program that takes ownership of the sync bus. This means that the clients can not change the sync directly and the owner has to make sure that all the time updates are pushed and are up to date to the clients. (This will not happen implicitly or automagically)

3. Because of how the API seems a lot more hierarchical, we would need to implement functionality so that we could setup a timing bus and drive if (if there isn't one already) or detect that it is already setup and connect as a client.

While I do think that the end result could be very nice, I don't think it is time well spent at this point as we already have issues finding people that uses the JACK sync functionality. So we don't really have any end users asking for this.

plugins/pipewire/PipeWireDevice.cpp
```diff
@@ -0,0 +1,376 @@
...
+void PipeWireDevice::PipeWireSynchronizer::seek(std::shared_ptr<IHandle> handle, double time)
+{
+	pw_time tm;
+	AUD_pw_stream_get_time_n(m_device->m_stream, &tm, sizeof(tm));
+	m_tick_start = tm.ticks;
+	m_seek_pos = m_timeline_pos = time;
+	{
+		std::unique_lock<std::mutex> lock(m_device->m_mixingLock);
+		m_device->m_clear_ringbuffer = true;
+	}
```
neXyon, line 51: Please remove the functionality to clear the ring buffer. The reason is that this can cause skips in sounds that are playing back but not being synchronized. This is especially bad for live sources like microphones as there would not even be any future data available for them.

plugins/pipewire/PipeWireDevice.h
```diff
@@ -0,0 +1,132 @@
...
+	{
+		PipeWireDevice* m_device;
+		int64_t m_tick_start = 0.0f;
+		double m_seek_pos = 0.0f;
+		double m_timeline_pos = 0.0f;
+
+	public:
+		PipeWireSynchronizer(PipeWireDevice* device);
+
+		void update_tick_start();
```
neXyon, line 54: Please rename to `camelCase`.

plugins/pipewire/PipeWireDevice.h
```diff
@@ -0,0 +1,132 @@
...
+
+	/**
+	 * The mixing ringbuffer and mixing data
+	 */
+	spa_ringbuffer m_ringbuffer;
+	Buffer m_ringbuffer_data;
+	bool m_clear_ringbuffer;
+	std::condition_variable m_mixingCondition;
+
+	AUD_LOCAL static void handle_state_changed(void* device_ptr, enum pw_stream_state old, enum pw_stream_state state, const char* error);
```
neXyon, line 90: `camelCase`

plugins/pipewire/PipeWireDevice.h
```diff
@@ -0,0 +1,132 @@
...
+	/**
+	 * Updates the ring buffers.
+	 */
+	AUD_LOCAL void updateRingBuffers();
+
+	/**
+	 * Mixes the next bytes into the buffer.
+	 * \param data The PipeWire device.
+	 */
+	AUD_LOCAL static void mix_audio_buffer(void* device_ptr);
```
neXyon, line 101: `camelCase`

plugins/pipewire/PipeWireDevice.h
```diff
@@ -0,0 +1,132 @@
...
+ */
+class AUD_PLUGIN_API PipeWireDevice : public SoftwareDevice
+{
+private:
+	class PipeWireSynchronizer : public DefaultSynchronizer
+	{
+		PipeWireDevice* m_device;
+		int64_t m_tick_start = 0.0f;
+		double m_seek_pos = 0.0f;
+		double m_timeline_pos = 0.0f;
```
neXyon, line 49: Just with the pulseaudio changes, I'm pretty sure it should be possible to do everything with just two variables here instead of three. But it's not critical to get this patch in.

neXyon (changes requested):
Great to hear that adding the ring buffer improved things. I would only request to remove the clearing of the ring buffer due to the stated reasons above. I understand that it slightly worsens the accuracy of the seeking but we should compensate with the same mechanism as described in the pulseaudio PR.

plugins/pipewire/PipeWireDevice.cpp
DarkDefender: I had to clear it as it would otherwise produce sound glitches when starting playback after pausing and seeking.
(Because old audio data would be played back before the new data could be queued)

neXyon: It's causing other issues though. I had to solve the same problem for Pulseaudio, please check how it's done there. In a nutshell: you just keep playing until the ring buffer is empty and are sure all samples from it have been played and then stop.

DarkDefender:
I think I've addressed all of your feedback.

I had to change around some more things as I noticed that you can't really query the stream playback state before you have actually got the callback to queue buffers. Which makes sense as you can't really get any accurate time data until you have actually started to queue audio data. IE you can't really ask "what is the audio delay" before you have actually started queuing/playing back data.

An other quirk I noticed and why I previously cleared the ringbuffer when seek was called:
The reason I'm getting audio glitches is that Blender seems to send "play" first and then send the seek command.

So it kinda looks like this: Play -> Seek to 2 sec
This of course makes it so that Audaspace has queued up outdated data which leads to audio glitches.
I'm looking into fixing this on the Blender side as it should seek and then play.
I think this will also fix the audio glitches I got with the pulseaudio backend.

I don't expect any more changes needed in audaspace to fix this. I just wanted to give you an update on "there is old audio still in the ringbuffer when starting to playback after pausing.
Turns out that the ringbuffer is properly cleared, it is just that Blender doesn't give Audaspace the correct playback position before starting to playback.

DarkDefender:
Note that I think we probably want to move the tick updates into the "process" callback function either way as this is makes sure that we don't start moving the playhead (synchronizer position) before any audio data actually has been sent to the sound system.

Later on we would modify this to take the ringbuffer delay into account so we would only start counting up when the queued data tied to the synchronizer has actually been queued up.

neXyon (approved):

neXyon:
Great! Thanks for your hard work on this! :)

PR #47, opened 2025-01-09
Add AAC container

aras-p:
You could already use Matroska container with AAC codec, but if someone wants raw .aac file ("ADTS stream"), there was no container option to pick.


plugins/ffmpeg/FFMPEGWriter.cpp
```diff
@@ -179,7 +179,7 @@ FFMPEGWriter::FFMPEGWriter(const std::string &filename, DeviceSpecs specs, Conta
 	m_input_samples(0),
 	m_deinterleave(false)
 {
-	static const char* formats[] = { nullptr, "ac3", "flac", "matroska", "mp2", "mp3", "ogg", "wav" };
+	static const char* formats[] = { nullptr, "ac3", "flac", "matroska", "mp2", "mp3", "ogg", "wav", "adts" };
```
neXyon, line 182: Why is it adts here and not aac like everywhere else?

neXyon: Ah well, apparently the muxer is called adts, so this is fine.

aras-p: Initially my wild guess was also "aac" until I found out that does not actually work... yes ffmpeg calls the muxer "adts".

PR #51, opened 2025-06-13
WIP: add time stretcher effect with rubberband

DoubleTrio:
This PR adds a TimeStretch effect using the Rubberband Library with the idea that this could be used for pitch correction in Blender. 

My implementation is probably way off and will definitely requires changes.

## Notes

- I added the Rubberband Library as a static library and created a new `external` directory for it
- Created `TimeStretch` and `TimeStretchReader` in their `fx` folder and I've tested it works in combination with other effects like `Loop` and `Delay` and works with `FileWriter::writeReader` too
- For using a TimeStretch effect on a 201 minute song that I loaded from a `File`, it took the Rubberband Library ~15 seconds to process in `Offline Mode`. 
- Currently, the stretcher is initialized in [TimeStretchReader.cp](https://github.com/neXyon/audaspace/compare/master...DoubleTrio:pitch?expand=1#diff-6692c70bc08447535b42e17dfba0264faace0ccd96efb6ef0a81e1e225bb86f9) to be in Offline Mode (should it be set to Realtime Mode?) and considerations for the different flags (listed [here](https://github.com/breakfastquay/rubberband/blob/default/rubberband/RubberBandStretcher.h)) probably needs to be discussed too


CMakeLists.txt
```diff
@@ -792,6 +793,27 @@ if(WITH_PYTHON)
...
+	list(APPEND HDR ${RUBBERBAND_HDR})
+
+	list(APPEND LIBRARIES rubberband_lib)
+	if (APPLE)
+		list(APPEND LIBRARIES  "-framework Accelerate")
+	endif()
+
+	add_library(rubberband_lib STATIC external/rubberband/single/RubberBandSingle.cpp)
+endif()
+
```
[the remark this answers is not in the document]
DoubleTrio: Thanks for the review/feedback! I'll address your comments in the upcoming days!

This is just me being completely new to adding libraries to a C++ project (and finding adding static libraries quick and easy) and being a bit new to modern C++. I didn't realize how outdated by C++ background was haha. 

I was also incorrect about the process time, it's actually ~8.20 seconds after I compiled everything in Release mode!

src/fx/TimeStretchReader.cpp
```diff
@@ -0,0 +1,204 @@
...
+	{
+		size_t need = m_stretcher.getSamplesRequired();
+		if(need == 0)
+			break;
+
+		len = need;
+
+		m_buffer.assureSize(len * samplesize);
+		buf = m_buffer.getBuffer();
+
```
DoubleTrio, line 115: Here, I don't need to add use a member variable, right? This is the only time `m_buffer` is used in the `TimeStretchReader` class, as I pretty much mirrored `LinearResampleReader.cpp`

src/fx/TimeStretchReader.cpp
```diff
@@ -0,0 +1,204 @@
...
+                                                           RubberBandStretcher::OptionPitchHighQuality | RubberBandStretcher::OptionEngineFiner)),
+                time_ratio)
+{
+	study();
+}
+
+void TimeStretchReader::study()
+{
+	Specs specs = m_reader->getSpecs();
+	int buffersize = AUD_DEFAULT_BUFFER_SIZE;
```
DoubleTrio, line 45: I wasn't sure about this line also, I used a buffer of 1024 as a default to read through the entire audio file for the `m_stretcher.study` method

DoubleTrio:
Next, I was thinking of adding a `crispness` option too which would probably look like

https://hg.sr.ht/~breakfastquay/rubberband-qt-example/browse/src/Processor.cpp?rev=tip starting at line `582`

Then lastly, where would I apply the TimeStretch effect? Because I'm think of perhaps of adding a boolean variable to indicate whether to use pitch correction or not in `SequenceEntry.cpp` and reusing the AnimateableProperty for pitch. I know that the pitch value is read in the `SequenceHandle` [here](https://github.com/neXyon/audaspace/blob/46036fa6749588c2a3f5c51e4f8937220f1ce45a/src/sequence/SequenceHandle.cpp#L213C1-L214C28 ), but wasn't sure when/how the effect could be applied.

DoubleTrio:
Oops, hold on, one of the audio files didn't work with the time stretcher now, I'll look into it tomorrow...

include/fx/TimeStretchReader.h
```diff
@@ -0,0 +1,102 @@
...
+
+	virtual void seek(int position);
+	virtual int getLength() const;
+	virtual int getPosition() const;
+
+  /**
+   * Retrieves the time \ratio for the stretcher.
+   * \return The current time ratio value.
+   */
+  float getRatio() const;
```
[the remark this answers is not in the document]
neXyon: Ah, I see. I just briefly checked the rubberband documentation last time and wrongly thought you can set the ratio between process() and study() for the next block but there it is also not allowed. Then yes, we will have to use the realtime mode! Checking the documentation now there are a few interesting hints here regarding the integration: https://breakfastquay.com/rubberband/integration.html For example there are recommended flags. Looks like for a general case the finer engine is always recommended, but potentially with a short window if the CPU resources are limited.

I would definitely not reuse the pitch AnimateableProperty for time stretching because users might want to use both functionalities at the same time. Rubberband can even be used to do both!

src/fx/TimeStretchReader.cpp
neXyon: The point of having buffers like this one as members is to not have to (re)allocate memory every time the read method is called as this is very inefficient and slows down the processing considerably. This is also why there is `assureSize` which only reallocates if the buffer is too small, which means the buffer might be bigger than the requested size, resulting in less memory allocation operations.

src/fx/TimeStretchReader.cpp
neXyon: When you switch to realtime and work with the sizes requested in the read method, this will become obsolete.

neXyon:
> I was wondering should the TimeStretchReader also store the entire output buffer of the pitch-corrected audio for other effects like Loop because it goes through the stretcher process twice (which already takes a while going through the stretcher process once)

No, the reader should work in realtime mode and not store any output actually.

> Next, I was thinking of adding a crispness option too which would probably look like
> 
> https://hg.sr.ht/~breakfastquay/rubberband-qt-example/browse/src/Processor.cpp?rev=tip starting at line 582

Suggestion: This could be added to the quality enum, resulting in various options.

> Then lastly, where would I apply the TimeStretch effect? Because I'm think of perhaps of adding a boolean variable to indicate whether to use pitch correction or not in SequenceEntry.cpp and reusing the AnimateableProperty for pitch. I know that the pitch value is read in the SequenceHandle [here](https://github.com/neXyon/audaspace/blob/46036fa6749588c2a3f5c51e4f8937220f1ce45a/src/sequence/SequenceHandle.cpp#L213C1-L214C28), but wasn't sure when/how the effect could be applied.

As written before, users might want both effects. So you should add it as another effect with its own AnimateableProperty!

include/fx/TimeStretchReader.h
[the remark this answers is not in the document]
DoubleTrio: Ahhh, I see! Hmm, perhaps then, I should change the TimeStretcher (change name to something else) effect so it can do both pitch-scaling and time-scaling.

src/devices/SoftwareDevice.cpp
```diff
@@ -878,12 +926,15 @@ std::shared_ptr<IHandle> SoftwareDevice::play(std::shared_ptr<IReader> reader, b
 	// pitch
 
 	std::shared_ptr<PitchReader> pitch = std::shared_ptr<PitchReader>(new PitchReader(reader, 1));
```
DoubleTrio, line 928: Alright, I think I can remove the PitchReader class since the TimeStretch class handles both pitch and timestretch now

src/sequence/SequenceHandle.cpp
```diff
@@ -212,6 +209,10 @@ void SequenceHandle::update(double position, float frame, float fps)
 	m_handle->setVolume(value);
 	m_entry->m_pitch.read(frame, &value);
 	m_handle->setPitch(value);
+
+	m_entry->m_time_stretch.read(frame, &value);
+	m_handle->setTimeStretch(value);
```
DoubleTrio, line 214: Then, this is where the time-stretch value is set in SequenceHandle

plugins/openal/OpenALDevice.cpp
```diff
@@ -437,6 +445,17 @@ bool OpenALDevice::OpenALHandle::setPitch(float pitch)
 	return true;
 }
 
+float OpenALDevice::OpenALHandle::getTimeStretch()
```
DoubleTrio, line 448: Hmm, can/should I do anything else here? I currently just have placeholders for the OpenALHandle as I added a setTimeStretch and getTimeStretch method to IHandle and this extends it

plugins/openal/OpenALDevice.cpp
```diff
@@ -437,6 +445,17 @@ bool OpenALDevice::OpenALHandle::setPitch(float pitch)
...
+{
+	float result = std::numeric_limits<float>::quiet_NaN();
+	return result;
+}
+
+bool OpenALDevice::OpenALHandle::setTimeStretch(float timeStretch)
+{
+	return false;
+}
+
```
DoubleTrio, line 458: Hmm, can/should I do anything else here? I currently just have placeholders for the OpenALHandle as I added a setTimeStretch and getTimeStretch method to IHandle and this extends it

src/devices/SoftwareDevice.cpp
```diff
@@ -878,12 +926,15 @@ std::shared_ptr<IHandle> SoftwareDevice::play(std::shared_ptr<IReader> reader, b
...
-	reader = std::shared_ptr<IReader>(pitch);
+
+	std::shared_ptr<TimeStretchReader> timeStretch = std::shared_ptr<TimeStretchReader>(new TimeStretchReader(reader, 1, 1, TimeStretchQualityOption::FASTEST));
+
+	reader = std::shared_ptr<IReader>(timeStretch);
```
DoubleTrio, line 932: Alright, I think I can remove the PitchReader class since the TimeStretch (which change later to something like Stretcher) class handles both pitch and timestretch now

DoubleTrio:
So now `AP_PITCH` affects the pitch without the sound-effect of time-stretching and `AP_TIME_STRETCH` handles it instead

Here's an example with the garbage sound with the code

```
  AnimateableProperty* prop = entries[0]->getAnimProperty(AP_PITCH);
  float pitch = 2.0;
  prop->write(&pitch, 3 * FPS, 1);
  float pitch2 = 0.5;
  prop->write(&pitch2, 7 * FPS, 1);
```

From 3-7 seconds, the music breaks completely, but sounds normal again (but pitch shifted down) 
[broken_music.wav.zip](https://github.com/user-attachments/files/20917142/pitch_test.wav.zip). 

Then here it is calling `m_stretcher.reset()` before setting the pitch scale (which fixes the issue)
[withreset.wav.zip](https://github.com/user-attachments/files/20917197/withreset.wav.zip)

It's more noticeable it this example with the flickering transition: 
[moon_flicker.wav.zip](https://github.com/user-attachments/files/20917225/moon_flicker.wav.zip)

I think it would be ideal to figure out a solution without calling `reset`  and I probably have to carefully look over the integration section again to make sure I'm handling everything correctly

include/fx/TimeStretchReader.h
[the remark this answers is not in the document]
neXyon: The correct naming of this is **time stretching and pitch scaling**. You have multiple options, one is to just keep the name, one is to add PitchScaling, or you can also add PitchScaling with `using` as an alternative name for the class, so it kinda has two and users can use the name for which they primarily use it. I'm fine with either of these options.

plugins/openal/OpenALDevice.cpp
neXyon: If you want, you can support this in the `OpenALDevice` too, by adding the reader to the handle (when necessary). But keeping this as is is fine too. If you keep it, please directly return the NaN, there is no need for the result variable here really.

include/fx/TimeStretch.h
```diff
@@ -0,0 +1,94 @@
...
+
+	// Crispness options correspond to https://breakfastquay.com/rubberband/usage.txt and https://hg.sr.ht/~breakfastquay/rubberband-qt-example/browse/src/Processor.cpp?rev=tip
+	// NOTE: These only apply when the R2 engine is used, that is when OptionEngineFaster is used
+	CRISP_0 = 1 << 2,
+	CRISP_1 = 1 << 3,
+	CRISP_2 = 1 << 4,
+	CRISP_3 = 1 << 5,
+	CRISP_4 = 1 << 6,
+	CRISP_5 = 1 << 7,
+	CRISP_6 = 1 << 8,
```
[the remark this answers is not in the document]
DoubleTrio: Will produce examples later!

DoubleTrio:
Alright, I've produced some examples [here](https://drive.google.com/drive/folders/1oP7mVow_dk1R86whsNVBaq-ihvF1pVc5?usp=sharing)! pit0-70 would correspond to pitch scale of 0.70x, while st2-00 would correspond to a time-stretch of 2.00x. Each file is labeled with the crisp quality + (high speed or high quality pitch) option as well as the process time at the end in seconds.

src/fx/TimeStretchReader.cpp
```diff
@@ -0,0 +1,275 @@
...
+
+	int channels = m_reader->getSpecs().channels;
+	int len;
+	sample_t* buf;
+
+	int available = m_stretcher->available();
+	while(available < length + m_dropAmount && !m_finishedReader)
+	{
+		// size_t need = m_stretcher->getSamplesRequired();
+		// Note for the V3 engine, the needed samples can be 0 sometimes...
```
DoubleTrio, line 61: I choose the block size for right now to be 1024, using `m_stretcher->getSamplesRequired()` led to the required samples to be 0 in the V3 engine (high quality pitch) which ended the loop too early before we could retrieve the necessary amount of output samples from the stretcher (I'll adjust/remove the comments later in the code to mention that it's actually faster to use 1024 as the length than always use `getSamplesRequired`)

src/fx/TimeStretchReader.cpp
```diff
@@ -0,0 +1,275 @@
...
+		// For now, choose the block size process size to be 1024. It's actually
+		len = 1024;
+
```
DoubleTrio, line 64: I choose the block size for right now to be 1024 (not sure if we want to change this to something else), using `m_stretcher->getSamplesRequired()` led to the required samples to be 0 in the V3 engine (high quality pitch) which ended the loop too early before we could retrieve the necessary amount of output samples from the stretcher (I'll adjust the comments later in the code to mention that it's actually faster to use 1024 as the length than always use `getSamplesRequired` since it often requests a small amount)

DoubleTrio:
> So now `AP_PITCH` affects the pitch without the sound-effect of time-stretching and `AP_TIME_STRETCH` handles it instead, which is a breaking change now. I was wondering how important maintaining backwards compatibility is.
> 
> As for the almost completed, but bugged TimeStretcherEffect...
> 
> Here's an example with the garbage sound with the code
> 
> ```
>   AnimateableProperty* prop = entries[0]->getAnimProperty(AP_PITCH);
>   float pitch = 2.0;
>   prop->write(&pitch, 3 * FPS, 1);
>   float pitch2 = 0.5;
>   prop->write(&pitch2, 7 * FPS, 1);
> ```
> 
> From 3-7 seconds, the music breaks completely, but sounds normal again (but pitch shifted down) [broken_music.wav.zip](https://github.com/user-attachments/files/20917142/pitch_test.wav.zip).
> 
> Then here it is calling `m_stretcher.reset()` before setting the pitch scale (which fixes the issue) [withreset.wav.zip](https://github.com/user-attachments/files/20917197/withreset.wav.zip)
> 
> It's more noticeable it this example with the flickering transition: [moon_flicker.wav.zip](https://github.com/user-attachments/files/20917225/moon_flicker.wav.zip)
> 
> I think it would be ideal to figure out a solution without calling `reset` and I probably have to carefully look over the integration section again to make sure I'm handling everything correctly

Huh, seems that this got resolved too and is now working too

include/fx/TimeStretchReader.h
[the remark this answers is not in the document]
DoubleTrio: I think I will opt for the second option :)

DoubleTrio:
Alright, I think it should good to go now, and addressed your comments. I had a couple more questions.
 
I was wondering... would I need to do anything (or anything else) for the Python bindings? I noticed that the constants are added in `PyAPI.cpp` and so I could add `AP_PITCH_SCALE`

```
PY_MODULE_ADD_CONSTANT(module, AP_LOCATION);
PY_MODULE_ADD_CONSTANT(module, AP_ORIENTATION);
``` 


For `SoftwareDevice.cpp`, would I also need to be put somewhere with `WITH_RUBBERBAND` since there is a dependency on `TimeStretchPitchScale.cpp`
https://github.com/neXyon/audaspace/blob/1e0989a304fb24299e112add7b4bb463e93dd1b7/CMakeLists.txt#L44

src/fx/TimeStretchPitchScaleReader.cpp
```diff
@@ -0,0 +1,285 @@
...
+	while(available < length + m_dropAmount && !m_finishedReader)
+	{
+		// size_t need = m_stretcher->getSamplesRequired();
+		// if (need == 0) break;
+
+		// Note for the V3 engine, the needed samples can be 0 sometimes and it breaks out of the loop too early before getting the necessary length for the buffer
+		// For now, choose the block size process size to be 1024.
+		// It's also actually faster too than to always use getSamplesRequired();
+
+		len = 1024;
```
neXyon, line 72: I think the number of samples you use here should depend on the length you need to output and the time stretch parameter. I.e., if you speed up time 2x, it should need 2x more input samples than you want output samples. Of course that's not a precise number which is why this needs to be a loop and you have to add more samples as necessary.

src/fx/TimeStretchPitchScaleReader.cpp
```diff
@@ -0,0 +1,285 @@
...
+void TimeStretchPitchScaleReader::read(int& length, bool& eos, sample_t* buffer)
+{
+	if(length == 0)
+		return;
+
+	if(m_pitchScale == 1.0 && m_timeRatio == 1.0)
+	{
+		m_reader->read(length, eos, buffer);
+		return;
+	}
```
neXyon, line 54: This will not work properly if the engine has been activated, since you might have fed it samples that are then skipped if the animation changes the values here to 1. But see my overall suggestion of the review which - if implemented - would make this code not necessary anymore as everything could be run through the engine if the effect is applied.

neXyon (commented):
Thanks for the updates! I've been thinking about how to ingegrate this best into Blender, also regarding the issues that we have with needing to implement this separately in OpenAL, consider rubberband availability in the SoftwareDevice and having to consider bypassing the expensive effect processing when this effect is attached to every sound even when unused. I think I came up with a better solution, but we have to see if we can get that to work.

The idea is rather than having this effect added to the Sequence*, Device and Handle classes, we just add it as a normal effect to the sound before it is added to the Sequence* on the Blender side. The problem we have to solve then is animation, because that is currently only done in Sequence* classes with the AnimateableProperty. My idea here would be to have a AnimateableTimeStretchPitchScale class next to the current one. That one contains two AnimateableProperty instances and the corresponding AnimateableTimeStretchPitchScaleReader can update the values of the underlying TimeStretchPitchScaleReader in every read call. In Blender we can then handle this effect like the equalizer which can be applied as a modifier to sound strips.

So far the theory of how this should work. Practically, I think it would be best to separate this PR into two or potentially even more. The first of these two adds the AnimateableTimeStretchPitchScale and -Reader classes including a C and Python API (as for other effect readers). This way we can get the foundation for all further changes in already. You then can (1) already try to integrate this effect into Blender just like the equalizer, just that the time stretch and pitch scale are not animateable. If you don't want to be able to animate these, we could be even done here. If you want to add the animation functionality though you can (2) add the AnimateableTimeStretchPitchScale* classes with a C (and potentially Python API, but I think AnimateableProperty doesn't have one yet, so you would need to add that too, if you want, i.e. that would still be optional) and make the values animateable in Blender.

What do you think about these suggestions?

I also listened to your samples. What do you think about them? I saw in the documentation that most of the options you chose for the crispness parameter don't  influence the output of the R3 (finer, high quality) engine. Sine the [integration guide](https://breakfastquay.com/rubberband/integration.html) (see Which options to use) sounds like R2 is mostly for percussive sounds and we cannot assume that, maybe we should just offer the R3 engine? It also sounds better to me in your samples, but might be subjective. Just based on the integration guide, I would suggest to have the following options: always use OptionProcessRealTime, OptionEngineFiner and OptionChannelsTogether; for lower quality/faster processing add OptionWindowShort or for higher quality add OptionPitchHighQuality (though that might not work well if the pitch shift is animated). Depending on whether the animateable class is used or not, we could add OptionPitchHighConsistency. As a separate option, we could add OptionFormantPreserved for better results with vocals. That's how I read the integration guide, what do you think?

neXyon:
Sounds good, but I would add the C and Python API for `TimeStretchPitchScale` in the first PR already or have one between 1 and 2. This allows you to already start working on the Blender integration and I think that would be the better order to do things as you will have something useful for Blender users faster which motivates better. :)

PR #52, opened 2025-07-14
Part 1: Add TimeStretchPitchScale effect

DoubleTrio:
Part 1 of splitting up https://github.com/neXyon/audaspace/pull/51, which is adding the TimeStretchPitchScale effect

Small things to do before is ready:

- [x] Remove references to padAmount and dropAmount for TimeStretchPitchScale
- [x]  Add formant preserving option similar to https://hg.sr.ht/~breakfastquay/rubberband-qt-example/browse/src/Processor.cpp?rev=tip
- [x] Make the number of input samples depend on the output and time-stretch

### DIFF
--- a/include/fx/TimeStretchPitchScale.h
+++ b/include/fx/TimeStretchPitchScale.h
@@ -1,0 +1,94 @@
+/*******************************************************************************
+ * Copyright 2009-2025 Jörg Müller
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ ******************************************************************************/
+
+#pragma once
+
+/**
+ * @file TimeStretchPitchScale.h
+ * @ingroup fx
+ * The TimeStretchPitchScale class.
+ */
+
+#include "fx/Effect.h"
+
+AUD_NAMESPACE_BEGIN
+
+enum StretcherQualityOption
+{
+	FASTEST = 1 << 0, // Use the high speed pitch speed
+	HIGH = 1 << 1,    // Use high quality pitch option
+
+	// Crispness options correspond to https://breakfastquay.com/rubberband/usage.txt and https://hg.sr.ht/~breakfastquay/rubberband-qt-example/browse/src/Processor.cpp?rev=tip
+	// NOTE: These really only apply when the R2 engine is used, that is when OptionEngineFaster is used, though window size does affect the R3 engine.
+	CRISP_0 = 1 << 2,
+	CRISP_1 = 1 << 3,
+	CRISP_2 = 1 << 4,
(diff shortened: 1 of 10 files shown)


include/fx/TimeStretchPitchScale.h
```diff
@@ -0,0 +1,106 @@
...
+
+	// Crispness options correspond to https://breakfastquay.com/rubberband/usage.txt and https://hg.sr.ht/~breakfastquay/rubberband-qt-example/browse/src/Processor.cpp?rev=tip
+	// NOTE: These really only apply when the R2 engine is used, that is when OptionEngineFaster is used, though window size does affect the R3 engine.
+	CRISP_0 = 1 << 2,
+	CRISP_1 = 1 << 3,
+	CRISP_2 = 1 << 4,
+	CRISP_3 = 1 << 5,
+	CRISP_4 = 1 << 6,
+	CRISP_5 = 1 << 7,
+	CRISP_6 = 1 << 8,
```
DoubleTrio, line 42: Hmm, I'm thinking about changing this so that crispness option is an integer and pitch options is a boolean just like in the example https://hg.sr.ht/~breakfastquay/rubberband-qt-example/browse/src/Processor.cpp?rev=tip on line 567. They also only delete the stretcher/ reinitialize if the window option has changed too unlike the current implementation. Mind if I make the changes the above two changes?

DoubleTrio:
Alright, I tested out applying the effect in Blender with hardcoded time-stretch and pitch-scale values, and it seems to work!

neXyon (commented):
Thanks for this PR! It looks good, that's the right amount of changes for a PR and we can merge this soon I think.

Unfortunately, I haven't had time yet to compile/run/test it. I will do that soon hopefully.

I made some comments, I think we still have to clearly define the options/interface still so that it is easy to use for users even if they are not audio pros. I think you can potentially also add a python interface already. It's basically just one method like the C interface.

DoubleTrio:
> Ah, this is where you got the crispiness from! I'm not sure if this parameter is ideal. If we follow the integration guide that I've also been mentioning in the other PR, we might not want to use the R2 engine at all and only use the R3 engine. I'm trying to think of this from a user perspective. As a user I would like to have meaningful options that allow me to achieve settings that I want and neither too many nor too few options. In that sense, I think a crispiness option that goes from 1-7 (0-6) is not ideal. My assumption is that in this example these values are there to test various possible configurations, i.e., that example is not really made for users, more for developers to try out/test/debug the library. I think we should stick with the combinations of options presented in the integration documentation, no?

I guess you are right, the `crispness` option is exactly the most intuitive from the user perspective. I think I'll adjust the configuration based on the integration guide then. 

> Also, I was too if the m_length, m_position, and the seek method were correct too

>> I'm not sure what you're trying to say here?

Ah, I think I was a bit confused by figuring out calculating `m_length` because I set back to 0 when the `seek` method is called. And the `m_position` variable now makes sense to me looking at the other example.

DoubleTrio:
I'll add these in!

neXyon:
Thank you! I merged the PR and also made some minor changes/fixes already. I'd also like to add your name to the AUTHORS file, if you want. Please tell me the name I should put in there. Your names here are Kace and DoubleTrio, the AUTHORS file contains full real names so far, but that's not a must.

I think it would be best to now focus on the Blender integration, without animation first as I wrote before. We need to get the dependency in. You can also rebase the other PR. I already set up a commit to push the changes of this PR into Blender's repository, but I would like to credit you there with your name as well, so let me also know which one I should use there please.

DoubleTrio:
Works like a charm! :)

WIP: https://projects.blender.org/blender/blender/pulls/143347

PR #53, opened 2025-07-17
Part 2: AnimatedTimeStretchPitchScale effect

DoubleTrio:
Part 2 of splitting up https://github.com/neXyon/audaspace/pull/51, which is adding the AnimatedTimeStretchPitchScale effect, which seems to work well based on testing on different audio clips!

Small things to do before is ready:

- [ ] Wait for https://github.com/neXyon/audaspace/pull/52 to be merged in and make changes accordingly
- [x] Add the C/Python API (anything else to add if needed?)
- [x] Should maybe make the AnimatedTimeStretchPitchScale have the AnimatableProperties, not the Reader

Other Questions:

- Should I move AnimateableProperty outside of the Sequence directory now that AnimatedTimeStretchPitchScale also uses it?
- Would the C/Python API for AnimateableProperty look nearly identical to the implementation to `AUD_Sequence_setAnimationData` and `AUD_SequenceEntry_setConstantRangeAnimationData` (should the API be specific to AnimatedTimeStretchPitchScale or should perhaps C++ templates should be used?)

```c++
AUD_API void AUD_Sequence_setAnimationData(AUD_Sound* sequence, AUD_AnimateablePropertyType type, int frame, float* data, char animated)
{
	AnimateableProperty* prop = dynamic_cast<Sequence *>(sequence->get())->getAnimProperty(static_cast<AnimateablePropertyType>(type));
	if(animated)
	{
		if(frame >= 0)
		{
			prop->write(data, frame, 1);
		}
	}
	else
	{
		prop->write(data);
	}
}

AUD_API void AUD_SequenceEntry_setConstantRangeAnimationData(AUD_SequenceEntry* entry, AUD_AnimateablePropertyType type, int frame_start, int frame_end, float* data)
{
	AnimateableProperty* prop = (*entry)->getAnimProperty(static_cast<AnimateablePropertyType>(type));
	prop->writeConstantRange(data, frame_start, frame_end);
}
```

Currently, I'm thinking something like (and I'm we can maybe reduce the repetitive code)

```c++
AUD_API void AUD_animatedTimeStretchPitchScale_setConstantRangeAnimationData(AUD_Sound* sound, AUD_AnimateablePropertyType type, int frame_start, int frame_end, float* data)
{
	AnimateableProperty* prop = std::dynamic_pointer_cast<AnimatedTimeStretchPitchScale>(*sound)->getAnimProperty(static_cast<AnimateablePropertyType>(type));
	prop->writeConstantRange(data, frame_start, frame_end);
}
```

- Blender's currently uses retiming keys which uses animates the playback speed using `AP_PITCH.` For pitch-correction toggle, animating the time-stretch value is needed and would need to be set to (`1/pitch_value`) and the `pitch_value` needs to be set to 1. What would be the best way to proceed, because my first idea was thinking of adding a way to `mute` the pitch (stop reading the animated pitch value and set it to 1). Or perhaps Blender should should store the AnimateableProperty buffer for pitch and then restore it once pitch-correction is turned off
- Then later (something on the Blender side), handling audio scrubbing

### DIFF
--- a/src/fx/AnimateableTimeStretchPitchScaleReader.cpp
+++ b/src/fx/AnimateableTimeStretchPitchScaleReader.cpp
@@ -51,9 +51,9 @@ void AnimateableTimeStretchPitchScaleReader::seek(int position)
 	double time = double(position) / double(m_reader->getSpecs().rate);
 	float frame = time * m_fps;
 
-	float timeRatio = m_timeStretch->readSingle(position);
+	float timeRatio = m_timeStretch->readSingle(frame);
 	setTimeRatio(timeRatio);
-	float pitchScale = m_pitchScale->readSingle(position);
+	float pitchScale = m_pitchScale->readSingle(frame);
 	setPitchScale(pitchScale);
 	TimeStretchPitchScaleReader::seek(position);
 }


src/fx/AnimateableTimeStretchPitchScaleReader.cpp
```diff
@@ -0,0 +1,94 @@
...
+		if(samplesRead <= 0 || eos)
+		{
+			break;
+		}
+
+		processed += samplesRead;
+	}
+	m_position += processed;
+	length = processed;
+}
```
DoubleTrio, line 63: Here's the read loop, I wasn't sure if this was the best approach as it has to loop through each sample at a time rather than reading a block of samples

include/fx/AnimateableTimeStretchPitchScaleReader.h
```diff
@@ -0,0 +1,78 @@
...
+/**
+ * This class reads from another reader and applies time-stretching and pitch scaling with support for animating both properties.
+ */
+class AUD_API AnimateableTimeStretchPitchScaleReader : public EffectReader
+{
+private:
+	/**
+	 * The animateable time stretch, pitch scale effect
+	 */
+	AnimateableTimeStretchPitchScale* m_timeStretchPitchScale;
```
neXyon, line 42: It should be possible to use the `AnimateableTimeStretchPitchScaleReader` without having to keep the `AnimateableTimeStretchPitchScale` alive, or even without ever allocating it.

This means that all the parameters you have in the `AnimateableTimeStretchPitchScale` need to be passed to this class. For the `AnimateableProperty` members, in order to share them and be able to update them, you shoud store them in an `std::shared_ptr`.

include/fx/AnimateableTimeStretchPitchScale.h
```diff
@@ -0,0 +1,111 @@
...
+private:
+	/**
+	 * The starting factor by which to adjust the pitch.
+	 */
+	double m_pitchScale;
+
+	/**
+	 * The starting factor by which to stretch or compress time.
+	 */
+	double m_timeRatio;
```
neXyon, line 48: These two should be removed as the `AnimateableProperty` members replace them.

include/fx/AnimateableTimeStretchPitchScaleReader.h
```diff
@@ -0,0 +1,78 @@
...
+
+	/**
+	 * The time stretch and pitch scale reader
+	 */
+	std::shared_ptr<TimeStretchPitchScaleReader> m_reader;
+
+	/**
+	 * The current position.
+	 */
+	int m_position;
```
neXyon, line 52: This member shouldn't be necessary.

src/fx/AnimateableTimeStretchPitchScaleReader.cpp
```diff
@@ -0,0 +1,81 @@
...
+
+		if(samplesRead <= 0 || eos)
+		{
+			break;
+		}
+
+		processed += samplesRead;
+	}
+	m_position += processed;
+	length = processed;
```
neXyon, line 62: I think this is too complex. You don't need to do any block processing here. If that is wanted by some user of the API, then they can do that from the outside by calling the read function with the appropriate amount of samples. I would simply set the properties based on the animated values and then forward the read call, i.e., just one run of the loop at full length. It should not be necessary to update the values too often, we also don't do it for pitch animation and it sounds fine.

src/fx/AnimateableTimeStretchPitchScaleReader.cpp
```diff
@@ -0,0 +1,81 @@
...
+}
+
+int AnimateableTimeStretchPitchScaleReader::getLength() const
+{
+	return m_reader->getLength();
+}
+
+int AnimateableTimeStretchPitchScaleReader::getPosition() const
+{
+	return m_position;
```
neXyon, line 78: You can just forward the call to the reader, then you don't have to store `m_position`. You could even get rid of most of these methods if you inherit from the `TimeStretchPitchScaleReader` class instead of having a member for it.

neXyon (changes requested):
Looks good already! Much simpler than having to change the devices, right? :)

bindings/C/AUD_Sound.cpp
```diff
@@ -67,6 +68,7 @@
 using namespace aud;
 
 #define AUD_CAPI_IMPLEMENTATION
+#include "AUD_Sequence.h"
```
neXyon, line 71: I assume this is necessary for `AUD_AnimateablePropertyType`? I'm not a fan of this dependence, but it's also not critical. But if that's the reason and you don't mind, we could move `AUD_AnimateablePropertyType` to AUD_Types.h.

bindings/C/AUD_Sound.h
```diff
@@ -420,6 +420,39 @@ extern AUD_API AUD_Sound* AUD_Sound_mutable(AUD_Sound* sound);
...
+    /**
+     * Time-stretches and pitch scales a sound with animation support
+     * \param sound The handle of the sound.
+     * \param timeRatio The factor by which to stretch or compress time.
+     * \param pitchScale The factor by which to adjust the pitch.
+     * \param quality The processing quality level of the stretcher.
+     * \param preserveFormant Whether to preserve the vocal formants for the stretcher.
+     * \return A handle of the time-stretched, pitch scaled sound.
+     */
+    extern AUD_API AUD_Sound* AUD_Sound_animateableTimeStretchPitchScale(AUD_Sound* sound, double timeRatio, double pitchScale, AUD_StretcherQuality quality,
```
neXyon, line 433: Since this is a C API and C doesn't have booleans by default, could you please change the type to char like in the other functions? Same for `AUD_Sound_timeStretchPitchScale` where I missed it in the last review please.

src/fx/AnimateableTimeStretchPitchScaleReader.cpp
```diff
@@ -0,0 +1,43 @@
...
+#include "fx/AnimateableTimeStretchPitchScaleReader.h"
+
+#include "IReader.h"
+
+AUD_NAMESPACE_BEGIN
+
+AnimateableTimeStretchPitchScaleReader::AnimateableTimeStretchPitchScaleReader(std::shared_ptr<IReader> reader, std::shared_ptr<AnimateableProperty> timeStretch,
+                                                                               std::shared_ptr<AnimateableProperty> pitchScale, StretcherQuality quality,
+                                                                               bool preserveFormant) :
+    TimeStretchPitchScaleReader(reader, 1.0, 1.0, quality, preserveFormant), m_timeStretch(timeStretch), m_pitchScale(pitchScale)
```
neXyon, line 28: In order to properly initialize the stretcher, I would read the value at the beginning of the property instead of setting them to 1 here, otherwise if it doesn't start at 1, you immediately have a jump. The same could be useful to do in the seek function, since it resets and reinitializes the stretcher.

include/fx/AnimateableTimeStretchPitchScale.h
```diff
@@ -0,0 +1,99 @@
...
+
+	/**
+	 * Returns the starting time stretch factor.
+	 */
+	double getTimeRatio() const;
+
+	/**
+	 * Returns the starting pitch scale factor.
+	 */
+	double getPitchScale() const;
```
neXyon, line 81: I'm not really sure how useful these methods are. When/what for would you use them?

include/fx/AnimateableTimeStretchPitchScale.h
```diff
@@ -0,0 +1,99 @@
...
+
+	/**
+	 * Returns whether formant preservation is enabled.
+	 */
+	bool getPreserveFormant() const;
```
neXyon, line 86: Since you added a getter for this, you might want to add one for the quality too? Same in the `TimeStretchPitchScale` class too, actually.

neXyon (changes requested):
Looks very good already. My comments are mostly minor things. Do you want to include the python API here as well? This might be a little more work, since you also need to add the AnimateableProperty as a new type. It might be better to have another PR for this in this case.

DoubleTrio:
Ah, I think I will include the Python API in a separate PR and see what I can do there :)
As for the `seek` function for the animated reader, are we assuming that it's constant, that is, it is the length of the reader * by the current time-stretch factor, or do we try to "estimate" how many samples in the input reader so that we seek there, or is it not really needed?

bindings/C/AUD_Sound.h
neXyon: You missed just this line here.

bindings/C/AUD_Sound.h
```diff
@@ -17,7 +17,6 @@
 #pragma once
 
 #include "AUD_Types.h"
-
```
neXyon, line 20: Nit: spurious change

include/fx/AnimateableTimeStretchPitchScaleReader.h
```diff
@@ -0,0 +1,70 @@
...
+ * @file AnimateableTimeStretchPitchScaleReader.h
+ * @ingroup fx
+ * The AnimateableTimeStretchPitchScaleReader class.
+ */
+
+#include "fx/AnimateableTimeStretchPitchScale.h"
+#include "fx/EffectReader.h"
+#include "fx/TimeStretchPitchScaleReader.h"
+
+using namespace RubberBand;
```
neXyon, line 29: Nit: not necessary in this file?

include/fx/AnimateableTimeStretchPitchScale.h
```diff
@@ -0,0 +1,104 @@
...
+	 */
+	StretcherQuality getStretcherQuality() const;
+
+	/**
+	 * Retrieves one of the animated properties of the entry.
+	 * \param type Which animated property to retrieve.
+	 * \return A pointer to the animated property, valid as long as the
+	 *         entry is.
+	 */
+	AnimateableProperty* getAnimProperty(AnimateablePropertyType type);
```
neXyon, line 99: Since the AnimateableProperty is a shared_ptr, this should also return a shared_ptr. You will need this for the Python API to work properly since all audaspace python objects refer to shared_ptr audaspace objects.

include/fx/AnimateableTimeStretchPitchScale.h
```diff
@@ -0,0 +1,104 @@
...
+
+	/**
+	 * Returns the time stretch factor at a position in samples.
+	 */
+	double getTimeRatio(float position) const;
+
+	/**
+	 * Returns the pitch scale factor at a position in samples.
+	 */
+	double getPitchScale(float position) const;
```
neXyon, line 81: I'm not sure these methods are actually necessary since you can already do the same through `getAnimProperty`. I don't have a strong opinion about this though. I would at least change the return type to float as the `AnimateableProperty` only supports float anyway.

src/fx/AnimateableTimeStretchPitchScaleReader.cpp
```diff
@@ -0,0 +1,62 @@
...
+#include "IReader.h"
+
+AUD_NAMESPACE_BEGIN
+
+float getValueAtPosition(std::shared_ptr<AnimateableProperty> prop, float position)
+{
+	float value;
+	prop->read(position, &value);
+	return value;
+}
```
neXyon, line 30: You could mark this function `static` (to hide it in the compilation unit) and `inline` (to mark it for optimization) or - going the other direction (making it more accessible) - you could add a `readSingle` method to `AnimateableProperty` (ideally with with an assert on `m_count` to ensure it's 1). Either way, I'd also use it in the read and seek methods.

neXyon (commented):
Just a few more comments, partially of stuff I overlooked last time. I think with another update it's ready to merge?

neXyon:
> Ah, I think I will include the Python API in a separate PR and see what I can do there :) As for the `seek` function for the animated reader, are we assuming that it's constant, that is, it is the length of the reader * by the current time-stretch factor, or do we try to "estimate" how many samples in the input reader so that we seek there, or is it not really needed?

So far the pitch animation also works by assuming it's constant at the position you're seeking to, so I think it is fine to have the same behavior. To more correctly seek, you would have to integrate the position and time stretch over the whole length from beginning to the seek position which is quite computationally intensive in comparison and takes more time the further back you seek. You can never do this 100 % accurately either, since it depends on the buffer length that you call the read function with (that's basically the sampling rate you would have to use for the integration). That said, if you are eager to try it, you're welcome to implement and profile it, to see how slow it really is in comparison.

> Also for the description of AnimateableProperty::read which looks like below
> 
> ```
> /**
> * Reads the properties value.
> * \param position The position in the animation in frames.
> * \param[out] out Where to write the value to.
> */
> ```
> 
> Would this have to be changed to include samples, since the AnimatedTimeStretchPitchScale is using it in samples currently

Ah, true, I just realized that this is happening right now. I think, we need to change this because the resolution of the `AnimateableProperty` should be way lower than per sample. Could you add a parameter/member to the classes to set the frame rate/fps just like for the sequence classes. This then allows users of the class to control the resolution and in Blender we can use the actual FPS which is also the resolution of the animation system there, i.e., you cannot use a higher resolution anyway. This also means that you will have to do the same `frame` calculation as the `SequenceReader` class.

DoubleTrio:
> > Ah, I think I will include the Python API in a separate PR and see what I can do there :) As for the `seek` function for the animated reader, are we assuming that it's constant, that is, it is the length of the reader * by the current time-stretch factor, or do we try to "estimate" how many samples in the input reader so that we seek there, or is it not really needed?
> 
> So far the pitch animation also works by assuming it's constant at the position you're seeking to, so I think it is fine to have the same behavior. To more correctly seek, you would have to integrate the position and time stretch over the whole length from beginning to the seek position which is quite computationally intensive in comparison and takes more time the further back you seek. You can never do this 100 % accurately either, since it depends on the buffer length that you call the read function with (that's basically the sampling rate you would have to use for the integration). That said, if you are eager to try it, you're welcome to implement and profile it, to see how slow it really is in comparison.

I'll try this in a separate branch maybe :)
I'll perhaps do the Python api first

DoubleTrio:
Hmm, I was wondering, how necessary it is creating a Python object for the AnimateableProperty (though I think it still be good to add if we want to initialize a sound by passing the AnimateableProperty in the constructor). I was looking at BinauralSound as an example in C++ and Python and it looks something like this

```c++
BinauralSound(std::shared_ptr<ISound> sound, std::shared_ptr<HRTF> hrtfs, std::shared_ptr<Source> source, std::shared_ptr<ThreadPool> threadPool, std::shared_ptr<FFTPlan> plan);
```

For the AnimateableTimeStretchPitchScale constructor, it just looks like this

```c++
AnimateableTimeStretchPitchScale(std::shared_ptr<ISound> sound, float timeStretch, float pitchScale, StretcherQuality quality, bool preserveFormant, float fps);
```

and initializes the AnimateableProperty in the constructor like so

```c++
AnimateableTimeStretchPitchScale::AnimateableTimeStretchPitchScale(std::shared_ptr<ISound> sound, float timeStretch, float pitchScale, StretcherQuality quality,
                                                                   bool preserveFormant, float fps) :
    Effect(sound),
    m_quality(quality),
    m_preserveFormant(preserveFormant),
    m_fps(fps),
    m_timeStretch(std::make_shared<AnimateableProperty>(1, timeStretch)),
    m_pitchScale(std::make_shared<AnimateableProperty>(1, pitchScale))
{
}
```

It might potentially be a good idea to have another constructor for the AnimateableTimeStretchPitchScale perhaps like below
```c++
AnimateableTimeStretchPitchScale(std::shared_ptr<ISound> sound, std::shared_ptr<AnimateableProperty> timeStretch,  std::shared_ptr<AnimateableProperty> pitchScale, StretcherQuality quality, bool preserveFormant, float fps);
```

Which then, it definitely makes sense to add a Python object for the AnimateableProperty (either way, I would love to give it a shot to add the AnimateableProperty type in Python!)

I might potentially be missing something though and just wanted to make sure. Otherwise, I can quickly add the Python API for AnimateableTimeStretchPitchScale and it wouldn't be too much work then.

src/fx/AnimateableTimeStretchPitchScaleReader.cpp
```diff
@@ -0,0 +1,61 @@
...
+
+void AnimateableTimeStretchPitchScaleReader::seek(int position)
+{
+	double time = double(position) / double(m_reader->getSpecs().rate);
+	float frame = time * m_fps;
+
+	float timeRatio = m_timeStretch->readSingle(position);
+	setTimeRatio(timeRatio);
+	float pitchScale = m_pitchScale->readSingle(position);
```
neXyon, line 56: You should also use the frame here!

neXyon (commented):
Looks good. One minor oversight in the seek function.

I still need to test it though, do you have some testing code that tests the functionality?

DoubleTrio:
Here's my general test file for everything. It's a bit messy though!

```c++
#include <iostream>
#include "file/File.h"
#include "file/FileWriter.h"
#include "fx/Delay.h"
#include "plugin/PluginManager.h"
#include "fx/TimeStretchPitchScale.h"
#include "fx/TimeStretchPitchScaleReader.h"
#include "fx/AnimateableTimeStretchPitchScale.h"
#include "fx/AnimateableTimeStretchPitchScaleReader.h"
#include "fx/Loop.h"
#include "fx/Pitch.h"
#include "fx/PitchReader.h"

#include "respec/LinearResample.h"

#include <chrono>
#include <iostream>

using namespace aud;


int main(int argc, char* argv[])
{
	if(argc != 5)
	{
		std::cerr << "Usage: " << argv[0] << " <filename>" << " <timeratio>" <<  "  <pitchscale>" <<  " <buffersize>" << std::endl;
		return 1;
	}


  double timeRatio = std::stod(argv[2]);
	double pitchScale = std::stod(argv[3]);
	int buffersize = std::stoi(argv[4]);
	float FPS = 30;



	
  PluginManager::loadPlugins("");
  
	std::shared_ptr<File> file = std::make_shared<File>(argv[1]);
  
	StreamInfo streamInfo = file->queryStreams()[0];






	auto start = std::chrono::high_resolution_clock::now();
  std::shared_ptr<AnimateableTimeStretchPitchScale> timeStretchPitchScale = std::make_shared<AnimateableTimeStretchPitchScale>(file, timeRatio, pitchScale, StretcherQuality::HIGH, false, FPS);
	
	std::shared_ptr<IReader> file_reader = file->createReader();
	

	DeviceSpecs outspecs;
	outspecs.specs = file_reader->getSpecs();
	outspecs.format = FORMAT_FLOAT32;

	std::shared_ptr<IWriter> writer = FileWriter::createWriter("stretchexample.wav", outspecs, CONTAINER_WAV, CODEC_PCM, 0);
	
	// Just make it large enough to handle any length...
	int totalsamples = 44100 * (60 * 30);

	auto timeReader = timeStretchPitchScale->createReader();


	
	auto time = timeStretchPitchScale->getAnimProperty(AP_TIME_STRETCH);
	float stretchFactor = 0.5;


		
	time->write(&stretchFactor, FPS * 10, 1);

	stretchFactor = 2;

	time->write(&stretchFactor, FPS * 15, 1);

	stretchFactor = 1;

	time->write(&stretchFactor, FPS * 20, 1);
	

		
	auto pitch = timeStretchPitchScale->getAnimProperty(AP_PITCH_SCALE);
	float pitchFactor = 1.2;
	stretchFactor = 0.5;
	pitch->write(&pitchFactor, FPS * 25, 1);
	time->write(&stretchFactor, FPS * 25, 1);
	FileWriter::writeReader(timeReader, writer, totalsamples, buffersize);
	auto end = std::chrono::high_resolution_clock::now();
	std::chrono::duration<double> elapsed = end - start;
	std::cout << "Time taken to write stretched file: " << elapsed.count() << " seconds\n";
}
```

PR #54, opened 2025-08-02
Part 3: Python API for AnimateableProperty and AnimatedTimeScalePitchScale 

DoubleTrio:
TODOs before ready:

- [ ] Implement methods and properties for python AnimateableProperty
- [ ] Implement methods and properties for python AnimatedTimeStretchPitchScale effect


neXyon:
Please rebase the branch on current master!

DoubleTrio:
Took a bit longer than I thought, but it's finished! Some things/notes I wasn't too sure of

- Should we add something for `getAnimProperty` in `PySequenceEntry` class?
- `{"animateableTimeStretchPitchScale_getAnimProperty", (PyCFunction) Sound_animateableTimeStretchPitchScale_getAnimProperty, METH_VARARGS, 
	M_aud_Sound_animateableTimeStretchPitchScale_getAnimProperty_doc}` - This is a very specific method to AnimateableTimeStretchPitchScale, and I'm not sure if there's a better way of doing this instead of just it inside of PySound.cpp


Also, here's my test python file!

```py
import aud
import numpy as np

prop = aud.AnimateableProperty(3)

value = np.array([0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9], dtype=np.float32)
prop.write(value, 0, 3)
print(prop.read(2))  # [0.7, 0.8, 0.9]
print(prop.count) # 3

prop2 = aud.AnimateableProperty(1, 3.0)

print(prop2.read(0)) # [3.0]
print(prop2.count) # 1
print(prop2.animated) # False

prop2.writeConstantRange(0.5, 3, 5)
print(prop2.read(4)) # [0.5]
print(prop2.animated) # True
print(prop2.read(1)) # [0.0]
print(prop2.readSingle(3)) # 0.5



value2 = np.array([111.1], dtype=np.float32)
prop2.write(value2, 20, 1)
print(prop2.read(20)) # [111.1]


print(dir(aud.Sound))
sound1 = aud.Sound.file("dt.mp3").timeStretchPitchScale(2.0, 1.0, 1, True)
sound2 = aud.Sound.file("dt.mp3").animateableTimeStretchPitchScale(2.0, 1.0, 1, True, 30.0)


prop3 = aud.Sound.animateableTimeStretchPitchScale_getAnimProperty(sound2, aud.AP_TIME_STRETCH)
print(prop3.read(0))
prop3.write(np.array([0.5], dtype=np.float32), 0, 1)
print(prop3.read(0))
```

bindings/python/PySound.cpp
```diff
@@ -1959,8 +2071,11 @@ static PyMethodDef Sound_methods[] = {
 #endif
 #ifdef WITH_RUBBERBAND
 	{"timeStretchPitchScale", (PyCFunction)Sound_timeStretchPitchScale, METH_VARARGS,
-	 M_aud_Sound_timeStretchPitchScale_doc
-	},
+	M_aud_Sound_timeStretchPitchScale_doc},
+  {"animateableTimeStretchPitchScale", (PyCFunction) Sound_animateableTimeStretchPitchScale, METH_VARARGS,
```
neXyon, line 2075: nit: indentation wrong?

bindings/python/PySound.cpp
```diff
@@ -1842,7 +1844,117 @@ Sound_timeStretchPitchScale(Sound* self, PyObject* args)
...
+
+    return (PyObject*)prop;
+}
+
+PyDoc_STRVAR(M_aud_Sound_animateableTimeStretchPitchScale_getAnimProperty_doc, ".. method:: animateableTimeStretchPitchScale_getAnimProperty(type)\n\n"
+                                                            "   Retrieves one of the animated properties of the sound (time-stretch or pitch-scaling).\n\n"
+                                                            "   :arg type: Which animated property to retrieve (AUD_AP_*).\n"
+                                                            "   :type type: int\n"
+                                                            "   :return: The created :class:`AnimateablePropertyP` object.\n"
+                                                            "   :rtype: :class:`AnimateablePropertyP`\n");
```
neXyon, line 1955: As you already noticed, the `animateableTimeStretchPitchScale_getAnimProperty` method doesn't really fit with the rest of the API. I think this should be changed so that two AnimateableProperty objects can be passed in to the `Sound_animateableTimeStretchPitchScale` method and the getter should be removed in order to fit with how everything else works here. You could even add such a constructor to the `AnimateableTimeStretchPitchScale` class.

bindings/python/PyAnimateableProperty.cpp
```diff
@@ -0,0 +1,353 @@
...
+	if(!PyArg_ParseTuple(args, "f", &position))
+		return nullptr;
+
+	int count = (*reinterpret_cast<std::shared_ptr<aud::AnimateableProperty>*>(self->animateableProperty))->getCount();
+	npy_intp dims[1] = {count};
+	PyObject* np_array = PyArray_SimpleNew(1, dims, NPY_FLOAT32);
+	if(!np_array)
+		return nullptr;
+
+	float* out = (float*) PyArray_DATA((PyArrayObject*) np_array);
```
neXyon, line 83: Please use `reinterpret_cast`, also for the other C style casts further down below.

bindings/python/PyAnimateableProperty.cpp
```diff
@@ -0,0 +1,353 @@
...
+	}
+
+	PyArrayObject* np_array = (PyArrayObject*) PyArray_FROM_OTF(array_obj, NPY_FLOAT32, NPY_ARRAY_IN_ARRAY);
+	if(!np_array)
+	{
+		PyErr_SetString(PyExc_TypeError, "data must be a numpy array of dtype float32");
+		return nullptr;
+	}
+
+	float* data_ptr = (float*) PyArray_DATA(np_array);
```
neXyon, line 162: I think for the python API you need to add checks if the np_array has the right size, i.e. the size is >= the required one. Also in the method below.

neXyon (commented):
Looks good already. One thing that you could look into is how hard it would be to make the AnimateableProperty act like a 2D np array directly, so that you could do something like this:

```python
prop[:, 0] = np.zeroes(prop.count)
```

DoubleTrio:
Hmm, a quick question. For reading values specifically, should we allow decimal slices, such as `print(prop[0.1:1.1:0.1])` will return the 10 animateable values?

neXyon:
> Hmm, a quick question. For reading values specifically, should we allow decimal slices, such as `print(prop[0.1:1.1:0.1])` will return the 10 animateable values? (Or is printing the slices not necessary?)
> 
> Writing to the array is more intuitive I think, just got to check the edge cases with slicing on my end now.

I think that's totally up to you, depending on how much functionality you want to support in the API. It should return an numpy ndarray, no? So you don't need to handle printing in any special way, no?

DoubleTrio:
> > Hmm, a quick question. For reading values specifically, should we allow decimal slices, such as `print(prop[0.1:1.1:0.1])` will return the 10 animateable values? (Or is printing the slices not necessary?)
> > Writing to the array is more intuitive I think, just got to check the edge cases with slicing on my end now.
> 
> I think that's totally up to you, depending on how much functionality you want to support in the API. It should return an numpy ndarray, no? So you don't need to handle printing in any special way, no?

Ah, it will return an numpy ndarray. I'm currently just adding a print statement just to test the results. Last thing to decide is whether to support negative indexes too.

bindings/python/PyAnimateableProperty.cpp
DoubleTrio: Just to clarify
```
prop = aud.AnimateableProperty(3)
```

Below would have the correct size
```
value = np.array([
    [0.1, 0.2, 0.3],
    [0.4, 0.5, 0.6],
    [0.7, 0.8, 0.9],
])
prop.write(value, 0, 3)
```

Also correct size
```
value = np.array([
    [0.1, 0.2, 0.3],
    [0.4, 0.5, 0.6],
    [0.7, 0.8, 0.9],
])
prop.write(value, 0, 2)
```

Incorrect size
```
value = np.array([
    [0.1, 0.2, 0.3],
    [0.4, 0.5, 0.6],
    [0.7, 0.8, 0.9],
])
prop.write(value, 0, 4)
```

Does the np array have to be 2 dimensional in this case, or it is also fine with the numpy array being 1D
```


value = np.array([ 0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9 ])
prop.write(value, 0, 3)
```

DoubleTrio: Opted for the last approach, but can change

neXyon (changes requested):
Tried it and it works nicely. I just have a few requests for changes:

First, in order to get it to compile, I had to include `<memory>` in `PyAnimateableProperty.cpp` for `std::shared_ptr`, please add that.

Second, I found the Python API hard to use. I think you should support keyword arguments for `Sound.timeStretchPitchScale` and `Sound.animateableTimeStretchPitchScale` and make pretty much all parameters optional (setting the currently non-optional ones to 1). The only parameter that should be required actually is the `fps` parameter for `animateableTimeStretchPitchScale`. To do that you will have to move it to the front and for consistency, please also do that for the C and C++ APIs.

Finally, for the `AnimateableProperty` API, the Python version could be simplified. Ideally, you would support the array access, you already started working on, but as I expected, this is not a quick addition an can be quite involved. That's why I suggested to do this as a separate PR in the first place. It's fine to not have that right now. One change you could make however is to simplify the write method to drop its count parameter and determine that from the length of the data array. If the length of the array is not a multiple of the property's count, you can either throw an error (preferred) or cut off the uneven part (i.e. rounding down when computing `count = array_length / prop.count`).

DoubleTrio:
By any chance, can you also share the Python file that you used to test too?

PR #58, opened 2025-09-02
Pitch animate seek test

DoubleTrio:


src/sequence/AnimateableProperty.cpp
```diff
@@ -237,6 +237,28 @@ float AnimateableProperty::readSingle(float position)
...
+	{
+		index = 0;
+	}
+
+	sample_t* buf = getBuffer();
+
+	return buf[index];
+}
+
+
```
neXyon, line 261: Instead of this method, I would add an explicit cast operator to buffer and then access the buffer directly, considering the range/size of the `Buffer` (`explicit operator const Buffer&() const { return *this; }`), otherwise you check these ifs for every sample during seek which is quite inefficient.

src/fx/AnimateableTimeStretchPitchScaleReader.cpp
```diff
@@ -45,14 +45,45 @@ void AnimateableTimeStretchPitchScaleReader::read(int& length, bool& eos, sample
...
+	double outputSamplePos = 0.0;
+
+	float ratio = 1.0f;
+	float lastRatio = 1.0f;
+
+	const double frameDuration = 1.0 / m_fps;
+	const double samplesPerFrame = frameDuration * sampleRate;
+	int frameIndex = 0;
+
+	while(outputSamplePos < position)
```
neXyon, line 67: Would be better to change this to a for loop and you can convert the position to an integer frame count to loop over (`int(frame)`) with `frameIndex`. Also: currently you're reading too far, i.e.,  `outputSamplePos` overshoots `position`. When you cast the frame count to int, rounding down, it will instead undershoot. See a comment further below how to get to the goal more precisely.

src/fx/AnimateableTimeStretchPitchScaleReader.cpp
```diff
@@ -45,14 +45,45 @@ void AnimateableTimeStretchPitchScaleReader::read(int& length, bool& eos, sample
...
+	{
+		ratio = m_timeStretch->readFrameSingle(frameIndex);
+
+		if(ratio <= 0.0f)
+			ratio = lastRatio;
+		else
+			lastRatio = ratio;
+
+		outputSamplePos += samplesPerFrame;
+		inputSamplePos += static_cast<int>(samplesPerFrame / ratio);
```
neXyon, line 77: You can increase the precision by keeping `inputSamplePos` as `double` during the computation and only cast it to int at the end when seeking the input reader.

src/fx/AnimateableTimeStretchPitchScaleReader.cpp
```diff
@@ -45,14 +45,45 @@ void AnimateableTimeStretchPitchScaleReader::read(int& length, bool& eos, sample
...
-	TimeStretchPitchScaleReader::seek(position);
-}
 
+	int inputSamplePos = 0;
+	double outputSamplePos = 0.0;
+
+	float ratio = 1.0f;
+	float lastRatio = 1.0f;
+
+	const double frameDuration = 1.0 / m_fps;
```
neXyon, line 63: This variable is not really needed, no? Just directly compute `samplesPerFrame` as `sampleRate / m_fps`.

src/fx/AnimateableTimeStretchPitchScaleReader.cpp
```diff
@@ -45,14 +45,45 @@ void AnimateableTimeStretchPitchScaleReader::read(int& length, bool& eos, sample
...
+			ratio = lastRatio;
+		else
+			lastRatio = ratio;
+
+		outputSamplePos += samplesPerFrame;
+		inputSamplePos += static_cast<int>(samplesPerFrame / ratio);
+
+		frameIndex++;
+	}
+
```
neXyon, line 81: As a final step, I'd call `readSingle` for the actual `position` and add `(position - total_frames * samplesPerFrame) / readSingle(...)` to `inputSamplePos` to account for the remainder.

neXyon (changes requested):
Since the frame approach seems to be better, I assume you want me to review this PR now :)

I commented in the code some improvements to get the accuracy and performance even better and then we can merge this!

I think we should even adopt this for the seeking with pitch animation in the SequenceReader class, but not in this PR!

DoubleTrio:
The samples are indeed much closer now (within the 3000-4000 sample range)
> Instead of this method, I would add an explicit cast operator to buffer and then access the buffer directly, considering the range/size of the Buffer (explicit operator const Buffer&() const { return *this; }), otherwise you check these ifs for every sample during seek which is quite inefficient.

For adding the explicit cast operator to AnimateableProperty, I got the warning
```
 warning: conversion function converting 'aud::AnimateableProperty' to its base class 'aud::Buffer' will never be used [-Wclass-conversion]
 ```

**Comment by neXyon** [comment]
> For adding the explicit cast operator to AnimateableProperty, I got the warning
> 
> ```
>  warning: conversion function converting 'aud::AnimateableProperty' to its base class 'aud::Buffer' will never be used [-Wclass-conversion]
> ```
> 
> It will also error if I try to cast to Buffer since the inheritance is private `class AUD_API AnimateableProperty : private Buffer`
> 
> Now, I wasn't sure if I should make it public instead of private or just add a getter for the Buffer in AnimateableProperty (which is what I currently have)

Ah, I see, the C++ standard doesn't allow that. I thought this would be a neat way to circumvent this. A `getBuffer` method is fine then!

**Comment by DoubleTrio** [comment]
I think it should be good now!

File: src/fx/AnimateableTimeStretchPitchScaleReader.cpp [hunk 6]
```diff
@@ -45,14 +45,50 @@ void AnimateableTimeStretchPitchScaleReader::read(int& length, bool& eos, sample
...
+	sample_t* buf = m_timeStretch->getBuffer();
+	for(int frameIndex = 0; frameIndex < totalFrames; frameIndex++)
+	{
+		ratio = buf[frameIndex];
+		if(ratio <= 0.0f)
+			ratio = lastRatio;
+		else
+			lastRatio = ratio;
+
+		outputSamplePos += samplesPerFrame;
```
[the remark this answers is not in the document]
DoubleTrio: I assumed that only

```
		outputSamplePos += samplesPerFrame;
```

gets removed from here

src/fx/AnimateableTimeStretchPitchScaleReader.cpp
```diff
@@ -45,14 +45,50 @@ void AnimateableTimeStretchPitchScaleReader::read(int& length, bool& eos, sample
...
+		inputSamplePos += samplesPerFrame / ratio;
+	}
+
+	double remainderSamples = position - outputSamplePos;
+	double remainderFrame = remainderSamples / samplesPerFrame;
```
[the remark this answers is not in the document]
DoubleTrio: Ah, I just did (frame - totalFrames) since totalFrame gets casted as an int

PR #63, opened 2025-11-17
Implemented a new Echo effect

AGoldenDragon:
This new PR introduces a completely new Echo effect.

The effect can be controlled using the familiar `Delay`, `Feedback`, and `Mix` parameters.

It is named `Echo` to avoid confusion with the existing `Delay` effect in Audaspace, which is used to delay a track’s start time.

**Video:**
https://github.com/user-attachments/assets/e390503c-0f98-42fd-addb-431450121ce0
**Sound used in the video:**
[Hello](https://freesound.org/people/gamer500/sounds/697809/) by [gamer500](https://freesound.org/people/gamer500/) | License: Creative Commons 0
(The video also demonstrates its implementation as a modifier for the base Blender branch.)

### DIFF
--- a/include/fx/Echo.h
+++ b/include/fx/Echo.h
@@ -1,30 +1,37 @@
 #pragma once
+
 /**
  * @file Echo.h
  * @ingroup fx
  * The Echo class.
  */
+
 #include "fx/Effect.h"
 #include <vector>
+
 AUD_NAMESPACE_BEGIN
+
 class AUD_API Echo : public Effect
 {
 private:
-    double m_delaySec;      /* Delay time in seconds */
-    double m_feedback;      /* Feedback amount */
-    double m_mix;           /* Wet/dry mix */
+    float m_delay;          /* Delay time in seconds */
+    float m_feedback;       /* Feedback amount */
+    float m_mix;            /* Wet/dry mix */
+    bool m_resetBuffer;     /* Whether to reset the delay buffer */
 
 public:
-    Echo(std::shared_ptr<ISound> sound, double delaySec, double feedback, double mix);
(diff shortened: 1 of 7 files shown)


include/fx/Echo.h
```diff
@@ -0,0 +1,31 @@
...
+ */
+#include "fx/Effect.h"
+#include <vector>
+AUD_NAMESPACE_BEGIN
+class AUD_API Echo : public Effect
+{
+private:
+    double m_delaySec;      /* Delay time in seconds */
+    double m_feedback;      /* Feedback amount */
+    double m_mix;           /* Wet/dry mix */
```
neXyon: Is there a reason you chose `double` as a datatype for these variables? I think single precision `float` should be more than enough here and again it would be more consistent with the rest of the engine and like more performant!

AGoldenDragon:
> There were a few more changes to be made, I pushed them already.
> 
> For the dry/wet mix though, shouldn't it be
> 
> ```c++
> sample_t outSample = inSample * (1.0f - m_mix) + (inSample + delayedSample * m_feedback) * m_mix;
> ```
> 
> instead of
> 
> ```c++
> sample_t outSample = inSample * (1.0f - m_mix) + delayedSample * m_mix;
> ```
> 
> as you want the whole effect output when the signal should be fully wet? Otherwise the original signal is missing and you hear only the echo?

I guess it's just a different style of the effect, but I agree that your version makes more sense.
By the way, if you could get it merged upstream into Blender’s base as soon as possible, that would be great, since I want to turn it into an Audio Modifier as soon as I can.
Thank you for everything🙃

PR #68, opened 2026-03-01
Audaspace: Add OpenAL capture support

phantomsoldierking:
Add basic audio capture support using the OpenAL capture API. CaptureDevice opens and manages an OpenAL input device, converts S16 samples to float32, and exposes sample querying and device enumeration. Includes spec normalization (mono/stereo, float32 output). Throws `DeviceException` when OpenAL is not available.

This is for the voiceover implementation in blender [Voiceover PR](https://projects.blender.org/blender/blender/pulls/154887) 
Ref:  [VSE: Voiceover #153779](https://projects.blender.org/blender/blender/issues/153779)


phantomsoldierking:
Was not sure about modifying `OpenALReader` so created a new `CaptureDevice`.

neXyon:
Please use `OpenALReader` which you already mentioned, you can certainly modify/extend it. The way you are implementing this at the moment is going against the plugin structure for dependencies that audaspace has. If you want we can discuss what changes you want to do and how to best do them code design wise, so that it fits with the design of the remaining library.

phantomsoldierking:
I have extended the `OpenALReader` as you suggested, hope this is correct.
Please let me know if any changes are required.
The intention was to be able to support multiple input sources for recording audio(for voiceover).
The complete pr for it is at [Voiceover PR](https://projects.blender.org/blender/blender/pulls/154887/files)

neXyon:
Thanks for the update, this looks much better already!

I would like to keep the API closer to the way the `DeviceManager` works: If I understand the code correctly, right now it's not possible to have multiple capture devices from multiple plugins. Just like for output devices it should be possible to register multiple input devices. Do you want to give this a go, or should I have a closer look?

phantomsoldierking:
Would be helpful if you could help out for other plugins(no Idea on `CoreAudio` and `WASAPI`).  
Please let me know if any changes are required in the `OpenAL` part.  I suppose the `DeviceManager` part need to be more like a map like in `m_factories`?

neXyon:
> Would be helpful if you could help out for other plugins(no Idea on `CoreAudio` and `WASAPI`).

Sure, that's not urgent at the moment and maybe other people want to help there.

> I suppose the `DeviceManager` part need to be more like a map like in `m_factories`?

Yes. And this is the part where I meant if I should help?

phantomsoldierking:
Oh! sorry, I have made the changes (used a map like output devices, with prefix `OpenAL - `) but no priority ordering.
Please check.

include/devices/DeviceManager.h
```diff
@@ -34,6 +35,13 @@ AUD_NAMESPACE_BEGIN
 class IDevice;
 class IDeviceFactory;
 class I3DDevice;
+class IReader;
+
+struct CaptureDeviceFactory
+{
+	std::vector<std::string> (*getDeviceNames)();
+	std::shared_ptr<IReader> (*openDevice)(const std::string& name, Specs specs, int buffersize);
+};
```
neXyon: Please turn this into an interface class like `IDeviceFactory` and name it `ICaptureDeviceFactory` in its own file.

src/devices/DeviceManager.cpp
```diff
@@ -119,4 +121,50 @@ std::vector<std::string> DeviceManager::getAvailableDeviceNames()
...
+	const std::string device = name.substr(separator_pos + separator.size());
+
+	auto it = m_capture_factories.find(backend);
+	if(it == m_capture_factories.end())
+		AUD_THROW(DeviceException, "The capture device couldn't be opened.");
+
+	if(it->second.openDevice == nullptr)
+		AUD_THROW(DeviceException, "The capture device couldn't be opened.");
+
+	return it->second.openDevice(device, specs, buffersize);
```
neXyon: When you turn `CaptureDeviceFactory` into an interface class that is then implemented by an `OpenALCaptureDeviceFactory`, you can store the name there like it is done in `OpenALDeviceFactory` and this method can be as simple as `DeviceManager::openDevice`. I also don't think it's necessary to capture errors like if there are no capture devices available or a wrong name was supplied.

neXyon (changes requested):
Thanks it looks great. I only would ask you to change the struct into an interface class that is then implemented for OpenAL to stay consistent with how it works for the output devices. It's okay to leave out priority for now.

phantomsoldierking:
Thank you for the review, Please check the changes.
Also forget to ask earlier, I had a doubt regarding the device list `OpenAL` lists all devices(I/O or pluged/unpluged) 
Is there a way to list/check only valid ones, like for example I have
```
➜  phantom ./audaspace-build/test_capture
Capture devices (7):
  [0] OpenAL - Monitor of Raptor Lake-P/U/H cAVS HDMI / DisplayPort 3 Output
  [1] OpenAL - Monitor of Raptor Lake-P/U/H cAVS Headphones
  [2] OpenAL - Monitor of Raptor Lake-P/U/H cAVS HDMI / DisplayPort 2 Output
  [3] OpenAL - Monitor of Raptor Lake-P/U/H cAVS HDMI / DisplayPort 1 Output
  [4] OpenAL - Raptor Lake-P/U/H cAVS Digital Microphone
  [5] OpenAL - Raptor Lake-P/U/H cAVS Stereo Microphone
  [6] OpenAL
Recording from: OpenAL - Raptor Lake-P/U/H cAVS Digital Microphone
Read 139776 frames total.
```
(but it should have been only the last 2)
Is it possible to filter or should we be listing all

neXyon (approved):
Thanks a lot for the updates, looks good now. I'll merge this and push it to blender as well.

PR #72, opened 2026-04-17
Upgrade SDL backend plugin from SDL2 to SDL3

Brainzman:
As described in the downstream [Blender PR #157521](https://projects.blender.org/blender/blender/pulls/157521) this PR upgrades the Audaspace SDL backend plugin from SDL2 to SDL3. Most changes were made following the [SDL3 audio migration guide](https://wiki.libsdl.org/SDL3/README-migration#sdl_audioh), notable changes include:

- The SDL Audio subsystem now needs to be explicitly initialized and closed
- Adapted the plugin to the new `SDL_AudioStream` paradigm, now opening the AudioDeviceStream and storing its reference in a new `m_stream` member
- Kept the mixing callback mechanism, adapted it to the new `SDL_AudioStreamCallback` signature
- CMake changes to obtain the new SDL3 library, using modern SDL3 CMake config targets

#### Changes local to the Audaspace repository
- `cmake/FindSDL2.cmake` was removed in favor of using CMake config files, which are now shipped by default with typical SDL3 packages (see the Archlinux [sdl3](https://archlinux.org/packages/extra/x86_64/sdl3/) package file list for example, shipping the SDL config files in `/usr/lib/cmake/SDL3`).

These changes can be tested directly in Blender by checking out [the downstream PR](https://projects.blender.org/blender/blender/pulls/157521) and compiling the SDL3 dependency via the Blender [dependency builder](https://developer.blender.org/docs/handbook/release_process/build/#make-deps).


neXyon (approved):
Thanks a lot!

PR #73, opened 2026-05-26
Fix AnimateableProperty::writeConstantRange resulting in garbage values when resampling

aras-p:
Fixes Blender bug where sometimes having retimed sound strips would crash (https://projects.blender.org/blender/blender/issues/155719). Blender PR: https://projects.blender.org/blender/blender/pulls/159091

The issue seems to be this:
1. Sound strips that are retimed use `AnimateableProperty::writeConstantRange` to write their section of pitch property values.
2. However, unlike `AnimateableProperty::write`, the `writeConstantRange` was not updating or handling "unknown" ranges of the animated values! So they ended up being completely uninitialized/garbage.
3. Now later on, when pitch value is read through `AnimateableProperty::read`, it does a cubic filter. Which means one of the filter taps can read from the "unknown" portion of the buffer, with whatever garbage value happened to be there.
4. As a result, effective pitch can end up being "anything", including values really close to zero, or values really close to infinities.
5. The JOS resample reader is not particularly safe against overflowing integers when it divides by ratios derived from this garbage pitch. So it ends up having "how much to read" being e.g. "4 billion" due to integer overflows.
6. Hilarity ensues!

From what I can tell, the problem is "just" that the `writeConstantRange`, when it was added in Audaspace 1.5 (db2ff5807765d), did not have the bit that fills in unknown parts of the buffer to have "something sensible".

### DIFF
--- a/include/sequence/AnimateableProperty.h
+++ b/include/sequence/AnimateableProperty.h
@@ -73,6 +73,7 @@ private:
 	AnimateableProperty& operator=(const AnimateableProperty&) = delete;
 
 	void AUD_LOCAL updateUnknownCache(int start, int end);
+	void AUD_LOCAL updateUnknownAfterWrite(int pos, int position, int count);
 
 public:
 	/**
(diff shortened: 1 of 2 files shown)


neXyon (approved):
Looks good, thanks!

PR #74, opened 2026-05-29
Fix sound.data() crash on non-existing file

aras-p:
Unlike all other bindings functions, Sound_data was not catching exceptions.

This is fixing Blender issue: https://projects.blender.org/blender/blender/issues/159049


neXyon (approved):
Thanks a lot!

